PR #80, opened 2024-09-17
Add Support for Edge to use Latest Build Images for Endpoint and Inference

honeytung:
This is the part 2 of the PRs that adds the `latest` tag to Docker Image so that `edge-endpoint` can always fetch the latest version.
Part 1 of the PR can be found here: https://github.com/positronix-ai/zuuul/pull/3969

### DIFF
--- a/deploy/balena-k3s/bastion/Dockerfile
+++ b/deploy/balena-k3s/bastion/Dockerfile
@@ -51,10 +51,6 @@ RUN arkade version && \
 RUN mkdir -p /app/edge-endpoint
 COPY . /app/edge-endpoint
 
-# Set environment variables for running cluster setup based on env var EDGE_INFERENCE_FLAVOR, defaults to CPU if not set
-ARG INFERENCE_FLAVOR_ARG="CPU"
-ENV INFERENCE_FLAVOR=${EDGE_INFERENCE_FLAVOR:-$INFERENCE_FLAVOR_ARG}
-ENV DEPLOYMENT_NAMESPACE="default"
 
 RUN chmod +x ./edge-endpoint/deploy/bin/cluster_setup.sh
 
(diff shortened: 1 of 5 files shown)


deploy/k3s/inference_deployment/inference_deployment_template.yaml
```diff
@@ -40,7 +40,7 @@ spec:
           maxUnavailable: 0  # Aim for no downtime during rollout
       containers:
       - name: inference-server
-        image: 767397850842.dkr.ecr.us-west-2.amazonaws.com/gl-tritonserver-test:674426bc8-main
+        image: 767397850842.dkr.ecr.us-west-2.amazonaws.com/gl-tritonserver-test:latest
```
tyler-romero, line 43: The tradeoff with doing this is that now whenever we push a new commit that builds this image, we will have a new :latest image. And these images are (currently) large (10-15Gb?) and take a long long time to pull to the edge device. Not a big concern at the moment but good to keep in mind that this isnt a free lunch.

tyler-romero (approved):

deploy/k3s/inference_deployment/inference_deployment_template.yaml
tomfaulhaber: What we really need here is a cicd pipeline matches all the versions together and some simpler tools for development. 

But I still think this is a step forward and we should do it.

deploy/balena-k3s/bastion/Dockerfile
```diff
@@ -51,10 +51,6 @@ RUN arkade version && \
 RUN mkdir -p /app/edge-endpoint
 COPY . /app/edge-endpoint
 
-# Set environment variables for running cluster setup based on env var EDGE_INFERENCE_FLAVOR, defaults to CPU if not set
```
tomfaulhaber, line 54: This means you're defaulting to GPU on Balena?

tomfaulhaber (approved):

deploy/balena-k3s/bastion/Dockerfile
honeytung: Yes, I think since our edge will be mostly running with GPU it is ok if we set it to use GPU by default. If we want to change that it is probably better if we moved this to `cluster_setup.sh` instead.

For now, RPi will need to set env var `INFERENCE_FLAVOR` to `CPU` in the balena dashboard.

PR #82, opened 2024-09-23
README and setup fixes/improvements

CoreyEWood:
Makes the READMEs give correct instructions for setting the endpoint up. Also improves some of the setup scripts to work more smoothly.

I tested the new setup process on a new user and it went smoothly.

In the future we might want to remove the code pertaining to the old docker setup (dockerfile, nginx config, etc.).


honeytung (approved):
LGTM! Thanks for the QoL improvements!

PR #87, opened 2024-09-24
Enabling EDGE_ONLY option

CoreyEWood:
Adds an option to only receive answers from the edge model when using `ask_async`. This code is temporary and will be removed once there is a better way to support this functionality.

### DIFF
--- a/app/api/routes/image_queries.py
+++ b/app/api/routes/image_queries.py
@@ -64,7 +64,7 @@ def handle_iqe_creation(
     app_state: AppState,
 ) -> ImageQuery:
     """
-    Handles the creation of an edge image query.
+    Handles the creation of an edge image query from the model results.
 
     :param detector_id: The string id of the detector to use
     :param results: A dict containing the results from the model.
--- a/app/api/routes/image_queries.py
+++ b/app/api/routes/image_queries.py
@@ -109,7 +109,7 @@ async def post_image_query(
         },
     )
 
-    edge_only = os.environ.get("EDGE_ONLY", "") == "ENABLED"
+    edge_only = bool(int(os.environ.get("EDGE_ONLY", 0)))
 
     # TODO: instead of just forwarding want_async calls to the cloud, facilitate partial
     #       processing of the async request on the edge before escalating to the cloud.
--- a/app/api/routes/image_queries.py
+++ b/app/api/routes/image_queries.py
@@ -167,12 +167,12 @@ async def post_image_query(
         confidence = results["confidence"]
 
         if (
+            edge_only or
             _is_confident_enough(
                 confidence=confidence,
                 detector_metadata=get_detector_metadata(detector_id=detector_id, gl=gl),
                 confidence_threshold=confidence_threshold,
             )
-            or edge_only
         ):
             if edge_only:
                 logger.info("EDGE_ONLY is enabled. The edge model's answer will be returned regardless of confidence.")
--- a/app/api/routes/image_queries.py
+++ b/app/api/routes/image_queries.py
@@ -109,57 +109,32 @@ async def post_image_query(
         },
     )
 
+    edge_only = os.environ.get("EDGE_ONLY", "") == "ENABLED"
+
+    # TODO: instead of just forwarding want_async calls to the cloud, facilitate partial
+    #       processing of the async request on the edge before escalating to the cloud.
+    _want_async = want_async is not None and want_async.lower() == "true"
+    if _want_async and not edge_only: # If edge-only mode is enabled, we don't want to make cloud API calls
+        return safe_call_api(
+            gl.submit_image_query,
+            detector=detector_id,
+            image=image,
+            wait=0,
+            patience_time=patience_time,
+            confidence_threshold=confidence_threshold,
+            human_review=human_review,
+            want_async=True,
+        )
+
     img_numpy = np.asarray(image)  # [H, W, C=3], dtype: uint8, RGB format
 
     edge_inference_manager = app_state.edge_inference_manager
     motion_detection_manager = app_state.motion_detection_manager
     require_human_review = human_review == "ALWAYS"
 
-    # TODO: instead of just forwarding want_async calls to the cloud, facilitate partial
-    #       processing of the async request on the edge before escalating to the cloud.
-    _want_async = want_async is not None and want_async.lower() == "true"
-    if _want_async:
-        edge_only = os.environ.get("EDGE_ONLY", "") == "ENABLED"
-        if not edge_only:
-            # Escalate to cloud api as normal
-            return safe_call_api(
-                gl.submit_image_query,
-                detector=detector_id,
-                image=image,
-                wait=0,
-                patience_time=patience_time,
-                confidence_threshold=confidence_threshold,
-                human_review=human_review,
-                want_async=True,
-            )
-
-        # Get the result of the edge model and return it
-        # NOTE: because this is temporary code, this skips handling motion detection
-
-        if not edge_inference_manager.inference_is_available(detector_id=detector_id):
-            raise RuntimeError("EDGE_ONLY is set to ENABLED, but edge inference is not available.")
-
-        logger.debug(
-            "EDGE_ONLY is set to ENABLED. This async request will not be escalated to the cloud and the "
-            "edge model's answer will be returned regardless of confidence. Motion detection will be ignored."
-        )
-
-        detector_metadata: Detector = get_detector_metadata(detector_id=detector_id, gl=gl)
-        results = edge_inference_manager.run_inference(detector_id=detector_id, img_numpy=img_numpy)
-
-        image_query = _handle_iqe_creation(
-            detector_id=detector_id,
-            results=results,
-            detector_metadata=detector_metadata,
-            patience_time=patience_time,
-            confidence_threshold=confidence_threshold,
-            app_state=app_state,
-        )
-
-        return image_query
-
     if not require_human_review and motion_detection_manager.motion_detection_is_available(detector_id=detector_id):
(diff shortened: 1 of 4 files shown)


app/api/routes/image_queries.py
```diff
@@ -54,6 +55,45 @@ async def validate_query_params_for_edge(request: Request, invalid_edge_params:
         )
 
 
+def handle_iqe_creation(
```
CoreyEWood, line 58: Abstracted this code into a helper because otherwise it would have been duplicated. Functionality should be the same.

app/api/routes/image_queries.py
```diff
@@ -108,26 +148,54 @@ async def post_image_query(
...
+            return safe_call_api(
+                gl.submit_image_query,
+                detector=detector_id,
+                image=image,
+                wait=0,
+                patience_time=patience_time,
+                confidence_threshold=confidence_threshold,
+                human_review=human_review,
+                want_async=True,
+            )
```
CoreyEWood, line 173: This preserves the same functionality as before when EDGE_ONLY is not set to ENABLED.

app/api/routes/image_queries.py
tyler-romero: Nice - we have a bunch of helper functions at the bottom of this file, can we colocate this with them, and also prepend an underscore to indicate that its a private helper function?

deploy/k3s/edge_deployment/edge_deployment.yaml
```diff
@@ -75,6 +75,10 @@ spec:
         # TODO: Once we have kubernetes-based tests, we can remove this feature flag.
         - name: DEPLOY_DETECTOR_LEVEL_INFERENCE
           value: "1"
+        # When this is set to ENABLED, ask_async calls to the edge endpoint will only return the edge model's answer,
+        # regardless of the confidence.
+        - name: EDGE_ONLY
+          value: "" # disabled by default
```
tyler-romero, line 81: Can we use an actual value that can be converted to a boolean here like "0" instead of an empty string?

CoreyEWood: Yeah for sure, should I also change the logic to interpret 0 / 1 values? Or keep it as is (where EDGE_ONLY must be "ENABLED").

app/api/routes/image_queries.py
```diff
@@ -108,26 +148,54 @@ async def post_image_query(
...
+
+        # Get the result of the edge model and return it
+        # NOTE: because this is temporary code, this skips handling motion detection
+
+        if not edge_inference_manager.inference_is_available(detector_id=detector_id):
+            raise ValueError("EDGE_ONLY is set to ENABLED, but edge inference is not available.")
```
tyler-romero, line 140: I think this is more appropriately a `RuntimeError`

app/api/routes/image_queries.py
```diff
@@ -108,26 +148,54 @@ async def post_image_query(
...
+
+        logger.debug(
+            "EDGE_ONLY is set to ENABLED. This async request will not be escalated to the cloud and the "
+            "edge model's answer will be returned regardless of confidence. Motion detection will be ignored."
+        )
+
+        detector_metadata: Detector = get_detector_metadata(detector_id=detector_id, gl=gl)
+        results = edge_inference_manager.run_inference(detector_id=detector_id, img_numpy=img_numpy)
```
tyler-romero, line 148: We shouldnt have to duplicate the call to run_inference here - I think this logic can be simplified.

As implemented, you're actually cutting motion detection out flow for this path.

The main difference with this new flag is that if it is enabled:
1. we bypass the want_async if statement
2. we return a value whether or not the edge inference model is confident enough.
3. if edge inference is not available, we throw an error

CoreyEWood: The reason I did it this way is because this is just temporary code for the demo, so I tried to avoid making modifications to the existing flow as much as possible. Leo also mentioned it might be simplest/most flexible to make the flag work only for `ask_async`, rather than all sdk calls. 

Since I'm not familiar with this code, I was worried there might be additional things to handle if the flag is incorporated more into the existing logic flow. I looked at the motion detection code and it seems like there's a chance it'll make a cloud API call - although looking at it again, it might not wait for a response, so maybe a delay wouldn't happen?

If we need motion detection for the demo and are okay with EDGE_ONLY being relevant to all api calls, I think it can be reworked as you suggested.

tyler-romero: I think my suggestion has a couple of strong points:
1. Fewer LOC changed for a demo (temporary code is sort of a myth IMO)
2. Less duplicated code, more consistent with existing behavior (its not really about the motion detection... until it is for some future demo or use case, then we're all like "why doesnt this work for motion detection anymore?")
3. Most importantly: makes it very obvious what the flag is actually doing differently - so contributors don't have to go line-by-line down the divergent code paths to figure it out for themselves.

If motion detection is making a call to the cloud, that is a separate issue we should address, rather than excluding motion detection altogether.

CoreyEWood: That makes sense to me, I'm down to rework it along those lines.

brandon-groundlight: I think I'm mostly with Tyler on this one. Skipping motion detection seems like a bug. I don't quite yet visualize what the full change is, so I'm not yet convinced that it will be fewer lines of code, but it certainly will be more correct.

Maybe a tangent, but I recall we were once in conversation with a potential partner who wanted to be able to use Groundlight to train models, but then they wanted to be able to cut out our cloud api costs once the model was trained. Things didn't work out at the time, but if someone like that came back to us I could see a business path where they keep buying hubs so long as the hubs can go fully offline at some point. They would be acutely interested in this behavior and then like Tyler said, the illusion of temporary code bursts

brandon-groundlight (commented):
Tyler made all the points and didn't leave enough points for me to make

app/api/routes/image_queries.py
```diff
@@ -202,6 +212,10 @@ async def post_image_query(
             }
         )
 
+        # Fail if edge inference is not available and edge-only mode is enabled
+        if edge_only:
+            raise RuntimeError("EDGE_ONLY is set to ENABLED, but edge inference is not available.")
```
CoreyEWood, line 214: I'm not 100% confident that this is the right place in the logic for this, so lmk if this seems wrong.

deploy/k3s/edge_deployment/edge_deployment.yaml
CoreyEWood: Bumping this question.

CoreyEWood (commented):
Reworked the logic to be more consistent with the previous flow. To make motion detection work with the EDGE_ONLY flag, I think some changes will be required, but I think it might be worth getting these changes out and then going back to motion detection later. The current state doesn't break motion detection, it just means that if EDGE_ONLY and motion detection are both enabled, we don't guarantee that no cloud API calls will be made.

app/api/routes/image_queries.py
honeytung: Does this just stops the program or you are handling the runtime error somewhere?

CoreyEWood: Right now it just stops the program. I wasn't sure what the best thing to do here is because there's the use case where the first time you submit an IQ to a detector, it will create the edge inference pod for it. That's why I put this below the `create_inference_deployment_record` call, so that the new pod will still get triggered even once the program stops. If you think there's a better way to handle this though I'm definitely down to change the behavior.

tyler-romero: I think from an API perspective this should look like a 500 error. IDK if fastapi is smart enough to do the translation to that on its own or if we need to explicitly handle it.

deploy/k3s/edge_deployment/edge_deployment.yaml
tyler-romero: Change the logic to interpret 0/1 values, like bool(int("1"))

app/api/routes/image_queries.py
```diff
@@ -108,10 +109,12 @@ async def post_image_query(
         },
     )
 
+    edge_only = os.environ.get("EDGE_ONLY", "") == "ENABLED"
```
tyler-romero, line 112: FWIW, if we did want to restrict this logic to only run when want_async is requested, it would be as simple as:
```
edge_only = (os.environ.get("EDGE_ONLY", "") == "ENABLED") and _want_async
```

tyler-romero (approved):
LGTM - would be good to check the behavior of that runtime error from a customer perspective though!

app/api/routes/image_queries.py
CoreyEWood: Just tested it and it does give a 500 response, with `Reason: Internal Server Error`.

honeytung: Yea for me I think it should automatically launch the pod and just revert to cloud API calls. Maybe use logger.warning to say that it is still waiting for inference pod to launch so it has to use cloud calls? 

But this way will still be slowing down the app if the inference pod is not ready. So I am not 100% sure if giving out 500 response to the client vs. just give cloud answer with an delay is a good solution.

tyler-romero: I dont think we should revert to calling the cloud, thats doing explicitly what the user said NOT to do.

tyler-romero: They could have good reason for not wanting to go to the cloud, such as cost

honeytung (approved):
LGTM! We need to note that the app needs to handle the exception raised when using `USE_EDGE` and the inference pod is not ready, otherwise, it is good to go!

app/api/routes/image_queries.py
CoreyEWood: Ah good point, that is easy actually. I think for now I'm fine with making all methods use the flag, but we could consider changing this in the future if we want to make it more restricted.

PR #90, opened 2024-09-27
Update k3s and nvidia install

tomfaulhaber:
I've updated our install sequence to use the default versions of k3s and the nvidia operator (directly from the helm chart).


deploy/bin/install-k3s-g4.sh
tyler-romero, line 1: While we're at it, should we rename this since it now works on more than just G4s?

tyler-romero (approved):
LGTM

deploy/bin/install-k3s-g4.sh
tomfaulhaber: Yes! Good catch. I'll change it to install-k3s-nvidia.sh.

PR #91, opened 2024-10-01
Enable per-detector edge-only mode

CoreyEWood:
Changes edge-only mode from being a global environment variable to being detector specific and configured in the `edge-config.yaml` file. 

Also makes `DetectorConfig`s be stored as a dict instead of a list.

See README for usage.

### DIFF
--- a/app/api/routes/image_queries.py
+++ b/app/api/routes/image_queries.py
@@ -135,7 +135,7 @@ async def post_image_query(
 
     if not require_human_review and motion_detection_manager.motion_detection_is_available(detector_id=detector_id):
         motion_detected = motion_detection_manager.run_motion_detection(detector_id=detector_id, new_img=img_numpy)
-        # TODO motion detection logic will likely need to be altered to work with the EDGE_ONLY flag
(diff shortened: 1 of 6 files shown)


app/api/routes/image_queries.py
```diff
@@ -251,6 +255,21 @@ async def get_image_query(
     return safe_call_api(gl.get_image_query, id=id)
 
 
+def _get_detector_config_by_id(detector_configs: list[DetectorConfig], detector_id: str) -> DetectorConfig | None:
```
tyler-romero, line 258: Could we just make detector_configs a set?

tyler-romero: I think this basically just looks like collapsing the list into a dict in `load_edge_config()`, but it:
a) removes the need to iterate over all detector configs on the critical path
b) is cleaner, since this method is basically just `my_dict.get(detector_id)`

CoreyEWood: That makes sense to me, definitely makes it easier to work with.

tyler-romero (approved):
LGTM thanks! Nice feature

PR #93, opened 2024-10-03
FastApi based inference-server

tyler-romero:
Switch us over to use a FastAPI-based edge inference-server for a few reasons:
1. Its simpler and more flexible
2. Its lighter weight and should reduce our inference-server image size
3. It should make it easy to support Jetsons, which is not supported when using triton w/ the python backend


app/core/edge_inference.py
```diff
@@ -1,19 +1,19 @@
...
 import logging
 import os
 import shutil
 import socket
 import time
 from typing import Dict, Optional
 
-import numpy as np
 import requests
-import tritonclient.http as tritonclient
```
tyler-romero, line 10: This file is where most of the changes are - swapping out triton's client and some changes to our model repository structure.

For now I just set us up to use `requests` to connect to the inference server, but an auto-generated client would be a good next step.

deploy/bin/build-push-edge-endpoint-image.sh
```diff
@@ -1,25 +1,46 @@
...
 
 
+set -ex
+
 # Ensure that you're in the same directory as this script before running it
 cd "$(dirname "$0")"
 
 TAG=$(./git-tag-name.sh)
-EDGE_ENDPOINT_IMAGE="edge-endpoint-test"  # temporary private repo name in GL_Public ECR
+# EDGE_ENDPOINT_IMAGE="edge-endpoint-test"  # v0.1.0 (triton inference server) compatible images
```
tyler-romero, line 31: Notably there is a switch-over happening here, even though the naming is terrible

deploy/k3s/edge_deployment/edge_deployment.yaml
```diff
@@ -40,7 +40,7 @@ spec:
       serviceAccountName: edge-endpoint-service-account
       initContainers:
       - name: database-prep
-        image: 767397850842.dkr.ecr.us-west-2.amazonaws.com/edge-endpoint-test:latest
+        image: &edgeEndpointImage 767397850842.dkr.ecr.us-west-2.amazonaws.com/edge-endpoint:latest
```
tomfaulhaber, line 43: TIL! I always get some nice reward for reading your PRs!

CoreyEWood (approved):
Very excited for this!! Looks great.

tyler-romero:
Totally agree tom, I pulled out the requests into intermediate functions and I think that looks a little nicer

PR #95, opened 2024-10-04
Support all detector modes

CoreyEWood:
Add support to the edge endpoint for all detector modes. Adds some basic tests to ensure responses are being parsed correctly.

### DIFF
--- a/app/api/routes/image_queries.py
+++ b/app/api/routes/image_queries.py
@@ -165,7 +165,6 @@ async def post_image_query(
         detector_metadata: Detector = get_detector_metadata(detector_id=detector_id, gl=gl)
         logger.debug(f"Local inference is available for {detector_id=}. Running inference...")
         results = edge_inference_manager.run_inference(detector_id=detector_id, image=image)
-        logger.info(f"On {detector_id}, got parsed edge model response in post_image_query. It was: {results=}")
         confidence = results["confidence"]
 
         if edge_only or _is_confident_enough(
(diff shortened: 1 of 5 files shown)


app/core/edge_inference.py
```diff
@@ -62,14 +62,48 @@ def parse_inference_response(response: dict) -> dict:
...
+    # Attempt to extract rois / text
+    if secondary_predictions is not None:
+        roi_predictions: list[list[dict]] | None = secondary_predictions.get("roi_predictions", None)
+        text_predictions: list[str] | None = secondary_predictions.get("text_predictions", None)
+        if roi_predictions is not None:
+            rois = roi_predictions[0]
+            for i, roi in enumerate(rois):
+                geometry = rois[i]["geometry"]
+                # TODO add validation to calculate x and y automatically
+                x = 0.5 * (geometry["left"] + geometry["right"])
```
robotrapta, line 96: Why are we calculating the centers here?

app/core/edge_inference.py
```diff
@@ -62,14 +62,48 @@ def parse_inference_response(response: dict) -> dict:
...
-    }
-    output_dict["label"] = "NO" if output_dict["label"] else "YES"  # map false / 0 to "YES" and true / 1 to "NO"
+
+    multi_predictions: dict = response.get("multi_predictions", None)
+    predictions: dict = response.get("predictions", None)
+    secondary_predictions: dict = response.get("secondary_predictions", None)
+
+    if multi_predictions is not None and predictions is not None:
+        raise ValueError("Got result with both multi_predictions and predictions.")
+    if multi_predictions is not None:
```
robotrapta, line 72: What's a "multi_prediction"?

robotrapta (approved):
Looks about right to me.

app/core/edge_inference.py
CoreyEWood: The `ImageQuery` model from the SDK requires `x` and `y` to be specified. Tyler thinks there's a way to get pydantic to do this automatically, but for now this seems easiest.

app/core/edge_inference.py
CoreyEWood: That's the field where multiclass/counting results are returned through.

app/core/edge_inference.py
tyler-romero: Yeah apparently Avi asked for the sdk to return the centers of the bboxes as well as left, top, right, bottom coordinates. Tim likes this too.

tyler-romero (approved):

app/api/routes/image_queries.py
```diff
@@ -186,13 +186,27 @@ async def post_image_query(
...
+                result_type = ResultTypeEnum.counting
+            elif mode == ModeEnum.MULTI_CLASS:
+                result_type = ResultTypeEnum.multi_classification
+            else:
+                raise ValueError(f"Got unrecognized detector mode: {mode}")
+
             image_query = create_iqe(
                 detector_id=detector_id,
+                result_type=result_type,
                 label=results["label"],
```
brandon-groundlight, line 203: This is going to change soon, you'll need to populate this field according to the mode type

app/core/utils.py
```diff
@@ -25,15 +28,15 @@ def create_iqe(
         created_at=datetime.utcnow(),
         query=query,
         detector_id=detector_id,
-        result_type=ResultTypeEnum.binary_classification,
+        result_type=result_type,
         result=BinaryClassificationResult(
```
brandon-groundlight, line 32: We'll need to use a different type here for counting modes

brandon-groundlight (commented):
I'm late to the party, this is great support for secondary predictions, but we'll eventually need to return to this to fully support the detector mode architecture

PR #96, opened 2024-10-05
Sync public model bucket to edge device upon inference server startup

tyler-romero:
In order to access groundlight base models on the edge device, we need to make them available at the same mount-point/file path as on prod. After discussing multiple ways to achieve this (explored s3fs-fuse as an alternative for on-demand fetching), we decided an s3 sync would be most prudent. 

This PR causes inference-server init containers to sync the s3 bucket, and uses a hostPath to persist the models on the edge-device.

On the node/device:
```
❯ ls /opt/groundlight/edge/pinamod-public/pinamod
groundingdino_swint_ogc_mmdet-822d7e9d.pth  yolox_s.pth
```


deploy/bin/cluster_setup.sh
```diff
@@ -180,6 +176,21 @@ if ! $K get pvc edge-endpoint-pvc; then
...
+if ! grep -q "^user_allow_other" /etc/fuse.conf; then
+    echo "Enabling user_allow_other in /etc/fuse.conf"
+    sudo sed -i 's/#user_allow_other/user_allow_other/' /etc/fuse.conf
+else
+    echo "user_allow_other is already enabled in /etc/fuse.conf"
+fi
+
+mkdir -p /opt/zudata/models/pinamod-public
+umount /opt/zudata/models/pinamod-public || true
+s3fs pinamod-artifacts-public:/pinamod /opt/zudata/models/pinamod-public -o ro -o allow_other -o use_cache=/tmp
```
tyler-romero, line 192: Doing this on the edge device doesnt feel great to me

pyproject.toml
```diff
@@ -10,7 +10,7 @@ APScheduler = "3.10.4"
 cachetools = "^5.3.1"
 fastapi = "^0.88.0"
 framegrab = "^0.5.0"
-groundlight = "^0.17.6"
+groundlight = "<0.18.0"
```
tyler-romero, line 13: edge is not yet compatible w/ newest sdk

configs/edge-config.yaml
```diff
@@ -35,6 +35,11 @@ detectors:
   # Blank id implies that no detectors have been configured to use motion detection
   # and/or local edge inference. The "default" templates are just examples of how one
   # might configure these features.
-  - detector_id: ''
-    motion_detection_template: "default"
+  - detector_id: 'det_2moISb6wdgxCgQPDu00KyFAM7GG'
```
tomfaulhaber, line 38: Are you intending to add real detector names here?

pyproject.toml
tomfaulhaber: Let's pin this to a specific version. We see this break on versions that should be compatible because the edge uses SDK internals directly.

tomfaulhaber (approved):
Just a couple of small comments

configs/edge-config.yaml
tyler-romero: Nope, thanks

PR #99, opened 2024-10-07
Support new modes from SDK changes

CoreyEWood:
Adds to previous code to fully support new mode structure. 

Multiclass is not supported for the time being but will be easy to add.


app/core/utils.py
```diff
@@ -1,47 +1,101 @@
...
+        )
+    elif mode == ModeEnum.COUNT:
+        if mode_configuration is None:
+            raise ValueError("mode_configuration for Counting detector shouldn't be None.")
+        count_mode_configuration = CountModeConfiguration(**mode_configuration)
+        max_count = count_mode_configuration.max_count
+        greater_than_max = False
+        if max_count is not None:
+            greater_than_max = result_value > max_count
+            result_value = max_count if greater_than_max else result_value
```
CoreyEWood, line 83: Is this correct logic for dealing with `max_count`?

app/core/utils.py
```diff
@@ -1,47 +1,101 @@
...
+        result_type = ResultTypeEnum.counting
+        result = CountingResult(
+            confidence=confidence,
+            source=source,
+            count=result_value,
+            greater_than_max=greater_than_max,
+        )
+    elif mode == ModeEnum.MULTI_CLASS:
+        raise NotImplementedError("Multiclass functionality is not yet implemented for the edge endpoint.")
+        # TODO add support for multiclass functionality.
```
CoreyEWood, line 93: I have code for this written, just need to clarify some implementation details once the mode is fully available.

app/api/routes/image_queries.py
```diff
@@ -188,16 +191,19 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
             else:
                 logger.debug(f"Edge detector confidence is high enough to return. {detector_id=}")
 
-            result_type, results = _mode_to_result_type(detector_metadata.mode, results)
-            patience_time = patience_time or constants.DEFAULT_PATIENCE_TIME
+            if patience_time is None:
+                patience_time = constants.DEFAULT_PATIENCE_TIME
+            if confidence_threshold is None:
+                confidence_threshold = detector_metadata.confidence_threshold
```
tyler-romero, line 197:
```suggestion
```
I moved this to the top of the file, since its needed for motdet as well :)

app/core/utils.py
tyler-romero: I think technically yes, because of how you implemented parse_inference_response().

Its a little messy though because we're performing the interpretation of a count result split over two different locations. When we come back to this for multiclass we should see about trying to clean that up a bit so there is one location where we do all the parsing and interpretation of results.

CoreyEWood: That makes sense, this was mostly me just dealing with things as I encountered them haha. Definitely can do a refactor in the next iteration.

test/database_manager/test_database_manager.py
```diff
@@ -152,8 +152,9 @@ def test_get_inference_deployments_raises_sqlalchemy_error(db_manager: DatabaseM
 def test_get_iqe_record(db_manager, database_reset):
     image_query: ImageQuery = create_iqe(
         detector_id=prefixed_ksuid("det_"),
-        result_type=ResultTypeEnum.binary_classification,
-        label="test_label",
+        mode=ModeEnum.BINARY,
```
tyler-romero, line 155: Can you add a test for creating a Count-mode IQE record as well?

tyler-romero: Just to make sure the DB is compatible with this change

CoreyEWood: Good idea, will do

PR #101, opened 2024-10-07
Edge only inference with cloud training

f-wright:
[[COM-1567](https://positronixcorp.atlassian.net/browse/COM-1567)] This PR adds an `edge_only_inference` mode which allows the model to escalate to the cloud for future training, while still always returning the edge inference answer for fast results. 

Tested with 3 binary detectors with
1. `edge_only` enabled -- nothing escalated to cloud
2. `edge_only_inference` enabled -- low confidence IQs escalated to cloud, edge answer returned
3. Neither `edge_only` nor `edge_only_inference` enabled -- low confidence edge IQs escalated to cloud and cloud answer returned. 
4. `edge_only` and `edge_only_inference` enabled -- pods do not launch, logs show the config validation error

Currently not adding many unit tests as they aren't set up to test `post_image_query`, and we want to get this change out quickly (discussed with Tyler).

This is a first step -- in the future we may want to add:
1. A task queueing system, instead of using FastAPI background tasks
5. Rate limiting for IQs sent to the cloud (either add an element of randomness for whether a query is sent to the cloud, or limit escalation to a certain number of queries over a period of time)


README.md
```diff
@@ -50,23 +50,32 @@ print(f"The answer is {image_query.result}")
 See the [SDK's getting started guide](https://code.groundlight.ai/python-sdk/docs/getting-started) for more info.
 
 ### Experimental: getting only edge model answers
-If you only want to receive answers from the edge model for a detector, you can enable edge-only mode for it. To do this, edit the detector's configuration in the [edge config file](./configs/edge-config.yaml) like so:
+If you only want to receive answers from the edge model for a detector, you can enable edge-only mode for it. This will prevent the edge endpoint from sending image queries to the cloud API. If you want fast edge answers regardless of confidence but still want the edge model to improve, you can enable edge-only inference for that detector. This mode will always return the edge model's answer, but it will also submit low confidence image queries to the cloud API for training.
```
CoreyEWood, line 53: I wonder if there's a different naming we could have for these different modes - I worry that the difference between "edge-only mode" and "edge-only inference" is pretty unclear to anyone external. Depending on how experimental/temporary these are, this might not matter, but if we expect external customers to use these at some point I think having more clear names might be helpful.

README.md
```diff
@@ -50,23 +50,32 @@ print(f"The answer is {image_query.result}")
...
+    motion_detection_template: "disabled"
+    local_inference_template: "default"
+    edge_only_inference: true
+
   - detector_id: 'det_abc'
     motion_detection_template: "default"
     local_inference_template: "default"
 ```
-In this example, `det_xyz` will have edge-only mode enabled because `edge_only` is set to `true`. If `edge_only` is not specified, it defaults to false, so `det_abc` will have edge-only mode disabled.
+In this example, `det_xyz` will have edge-only mode enabled because `edge_only` is set to `true`. `det_ijk` will have edge-only inference enabled because `edge_only_inference` is set to `true`. If `edge_only` or `edge_only_inference` are not specified, they defaults to false, so `det_abc` will have edge-only mode disabled. Only one of `edge_only` or `edge_only_inference` can be set to `true` for a detector.
```
CoreyEWood, line 72: Just a tiny typo but this says "they defaults to false".

README.md
```diff
@@ -50,23 +50,32 @@ print(f"The answer is {image_query.result}")
...
 
 With edge-only mode enabled for a detector, when you make requests to it, you will only receive answers from the edge model (regardless of the confidence). Additionally, note that no image queries submitted this way will show up in the web app or be used to train the model. This option should therefore only be used if you don't need the model to improve and only want fast answers from the edge model.
 
-If edge-only mode is enabled on a detector and the edge inference model for that detector is not available, attempting to send image queries to that detector will return a 500 error response.
+With edge-only inference enabled for a detector, when you make requests to it, you will only receive answers from the edge model (regardless of the confidence). However, image queries submitted this way with confidences below the threshold will be used to train the model. This option should be used when you want fast edge answers regardless of confidence but still want the model to improve.
```
CoreyEWood, line 76: Maybe to be more clear this should say something like, "However, when the edge model makes a prediction with confidence below the threshold, that image query will also be escalated to the cloud and used to train the model"?

test/core/test_configs.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+
+from app.core.configs import DetectorConfig
+
+
+def test_detector_config():
```
CoreyEWood, line 6: Awesome to have this test, maybe it would be good to name it something more specific like "test_detector_config_with_both_edge_modes" in case we want to add more detector config tests in the future?

app/api/routes/image_queries.py
```diff
@@ -207,6 +225,13 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
                 text=results["text"],
             )
             app_state.db_manager.create_iqe_record(iq=image_query)
+
+            if edge_only_inference and not _is_confident_enough(
+                confidence=confidence,
+                confidence_threshold=confidence_threshold,
+            ):
+                logger.info("Escalating to the cloud API server for future training due to low confidence.")
+                background_tasks.add_task(safe_call_sdk, gl.ask_async, detector=detector_id, image=image)
```
CoreyEWood, line 234: This is great! Love how simple this is.

CoreyEWood (approved):
This looks great overall! Awesome functionality to have. I just left a few comments about making sure things aren't confusing if external users are going to be using the edge-only modes.

As an additional thought, I wonder if we should change the default `refresh_rate` so that it looks for new model binaries for frequently - the current default of 120 seconds is pretty slow, and users might want to see improvements more quickly if they're sending a lot of queries.

README.md
f-wright: Oops thank you!

README.md
f-wright: I agree that it's confusing. Maybe I could change them to `edge_only_no_training` and `edge_only_cloud_training`? That gets a little wordy, but might be more clear. Definitely open to naming suggestions, wasn't sure what to call it.

brandon-groundlight (approved):
Looks good to me. I do think deduping the images will become important in the near future, but for internal demos this should be great

PR #106, opened 2024-10-08
Add timestamps to inference-model-updater logs

CoreyEWood:
Previously, logs in the inference-model-updater container of the edge-endpoint pod didn't have timestamps. Now they do! This makes it easier to know when the next update is going to happen, and generally to debug model updating.

I'm not an expert on how python logging setup works, so let me know if this should be done in a different way.

### DIFF
--- a/app/model_updater/update_models.py
+++ b/app/model_updater/update_models.py
@@ -9,15 +9,18 @@ from app.core.database import DatabaseManager
 from app.core.edge_inference import EdgeInferenceManager, delete_old_model_versions
 from app.core.kubernetes_management import InferenceDeploymentManager
 
-log_level = os.environ.get("LOG_LEVEL", "INFO").upper()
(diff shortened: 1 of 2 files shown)


app/model_updater/update_models.py
```diff
@@ -9,15 +9,18 @@
 from app.core.edge_inference import EdgeInferenceManager, delete_old_model_versions
 from app.core.kubernetes_management import InferenceDeploymentManager
 
-log_level = os.environ.get("LOG_LEVEL", "INFO").upper()
-logging.basicConfig(level=log_level)
+LOG_LEVEL = os.environ.get("LOG_LEVEL", "INFO").upper()
+logging.basicConfig(
```
CoreyEWood, line 13: Used the same format as is defined in app/main.py.

tomfaulhaber (approved):
lgtm

PR #109, opened 2024-10-10
Increase nginx worker connections

CoreyEWood:
Increases worker connections to 256. We could probably go even higher, but this should be a fine number for now.


tyler-romero (approved):
LGTM

PR #110, opened 2024-10-11
Rip out Motion Detection

tyler-romero:
Remove motion-detection, which is messy, adds complexity, and is not compatible with multiple workers. Edge-inference is intended to fill the same gap on the edge-endpoint. Really motion-detection should be implemented as close to the camera as possible (that is, before even sending requests to the edge-endpoint).


Dockerfile
```diff
@@ -86,7 +86,7 @@ COPY --from=production-dependencies-build-stage ${APP_ROOT}/configs/nginx.conf /
 # Remove default nginx config
 RUN rm /etc/nginx/sites-enabled/default
 
-CMD nginx && poetry run uvicorn --workers 1 --host 0.0.0.0 --port ${APP_PORT} --proxy-headers app.main:app
+CMD nginx && poetry run uvicorn --workers 8 --host 0.0.0.0 --port ${APP_PORT} --proxy-headers app.main:app
```
tyler-romero, line 89: Now that motion detection (which was not compatible with multiple workers) is gone, increase worker count to 8

CoreyEWood (approved):
Yay!

app/api/routes/image_queries.py
```diff
@@ -224,18 +160,26 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
                 rois=results["rois"],
                 text=results["text"],
             )
-            app_state.db_manager.create_iqe_record(iq=image_query)
+            background_tasks.add_task(app_state.db_manager.create_iqe_record, iq=image_query)
```
tyler-romero, line 163: Moving the db transaction to after the response is returned.
is this even worth it? need to benchmark.

PR #113, opened 2024-10-16
Add TTL cache for edge inference ready checks

CoreyEWood:
Before the edge endpoint proxy submits an IQ to the inference server, it checks for readiness. If many IQs are being processed, this results in a lot of checks. This PR adds a TTL cache to the readiness check that will cache the result for 30 seconds.

### DIFF
--- a/app/core/edge_inference.py
+++ b/app/core/edge_inference.py
@@ -19,7 +19,7 @@ logger = logging.getLogger(__name__)
 
 # Simple TTL cache for is_edge_inference_ready checks to avoid having to re-check every time a request is processed.
 # This will be process-specific, so each edge-endpoint worker will have its own cache instance.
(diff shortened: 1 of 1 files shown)


app/core/edge_inference.py
```diff
@@ -16,7 +17,12 @@
 
 logger = logging.getLogger(__name__)
 
+# Simple TTL cache for is_edge_inference_ready checks to avoid having to re-check every time a request is processed.
+# This will be process-specific, so each edge-endpoint worker will have its own cache instance.
+ttl_cache = TTLCache(maxsize=10, ttl=30)
```
tyler-romero, line 22:
```suggestion
ttl_cache = TTLCache(maxsize=128, ttl=30)
```

tyler-romero (approved):
LGTM! Thanks!

PR #115, opened 2024-10-18
Add rate limit to background escalations

f-wright:
This PR adds a rate limit to detectors running background escalations to the cloud. The default time to wait between escalations will be 2 seconds, but we'll allow customers to set that for themselves. If they set too small of a limit (or a limit of 0), they'll still be effectively limited by our usual throttling. 

Unit test coverage of `image_queries.py` is still a little lacking, and that's a larger fix than I think makes sense to do in this PR. I tested out the new option with the demo code.

### DIFF
--- a/app/api/routes/image_queries.py
+++ b/app/api/routes/image_queries.py
@@ -160,7 +160,7 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
             if not disable_cloud_escalation and not is_confident_enough:
                 # Only escalate if we haven't escalated on this detector too recently
                 if app_state.edge_inference_manager.escalation_cooldown_complete(detector_id=detector_id):
-                    logger.debug(
+                    logger.info(
                         f"Escalating to cloud due to low confidence: {ml_confidence} < thresh={confidence_threshold}"
                     )
                     background_tasks.add_task(
@@ -173,7 +173,7 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
                         human_review=human_review,
                     )
                 else:
-                    logger.debug(
+                    logger.info(
                         f"Not escalating to cloud due to rate limit on background cloud escalations: {detector_id=}"
                     )
 
--- a/app/api/routes/image_queries.py
+++ b/app/api/routes/image_queries.py
@@ -156,24 +156,26 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
             )
             app_state.db_manager.create_iqe_record(image_query)
 
-            # escalate after returning edge prediction if enabled and we haven't escalated on this detector too recently
-            if (
-                not disable_cloud_escalation
-                and not is_confident_enough
-                and app_state.edge_inference_manager.escalation_cooldown_complete(detector_id=detector_id)
-            ):
-                logger.debug(
-                    f"Escalating to cloud due to low confidence: {ml_confidence} < thresh={confidence_threshold}"
-                )
-                background_tasks.add_task(
-                    safe_call_sdk,
-                    gl.ask_async,
-                    detector=detector_id,
-                    image=image_bytes,
-                    patience_time=patience_time,
-                    confidence_threshold=confidence_threshold,
-                    human_review=human_review,
-                )
+            # Escalate after returning edge prediction if escalation is enabled and we have low confidence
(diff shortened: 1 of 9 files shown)


app/core/edge_inference.py
```diff
@@ -111,24 +111,39 @@ class EdgeInferenceManager:
     INFERENCE_SERVER_URL = "inference-service:8000"
     MODEL_REPOSITORY = MODEL_REPOSITORY_PATH
 
-    def __init__(self, config: Dict[str, LocalInferenceConfig] | None, verbose: bool = False) -> None:
+    def __init__(
+        self,
+        inference_configs: Dict[str, LocalInferenceConfig] | None,
+        edge_config: Optional[RootEdgeConfig] = None,
```
f-wright, line 117: We have a bit of a weird set up right now where we use a config as a detector arguments in `inference_config`, but specific values for `always_return_cloud_prediction` and `disable_cloud_escalation`.

We want the new `min_time_between_escalations` field to be configurable on a detector level, and I think it makes sense to keep it with `always_return_edge_prediction` and `disable_cloud_escalation` since it's only valid with a particular configuration of those fields. However, the best place I saw to keep track of the time of last escalation for a detector is in `EdgeInferenceManager` -- hence adding the `edge_config` input. 

I think it might be better to get rid of the `local_inference_template` in the detector config entirely, and just add the fields from it to the detector config. Alternatively, we could move all three of the new detector config fields into the `local_inference_template`. That reorganization felt unrelated to this PR, but happy to do it if others agree it should be done.

app/api/routes/image_queries.py
```diff
@@ -156,19 +156,26 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
...
-                    image=image_bytes,
-                    patience_time=patience_time,
-                    confidence_threshold=confidence_threshold,
-                    human_review=human_review,
-                )
+            # Escalate after returning edge prediction if escalation is enabled and we have low confidence
+            if not disable_cloud_escalation and not is_confident_enough:
+                # Only escalate if we haven't escalated on this detector too recently
+                if app_state.edge_inference_manager.escalation_cooldown_complete(detector_id=detector_id):
+                    logger.info(
```
CoreyEWood, line 163: I think we might want the two logs here to be debug level - this one might make sense to keep as info, but the 'not escalating to cloud' one could produce a lot of messages if the rate of requests was high.

app/core/edge_inference.py
```diff
@@ -111,21 +111,36 @@ class EdgeInferenceManager:
...
-        if config:
-            self.inference_config = config
+
+        # Last time we escalated to cloud for each detector
+        self.last_escalation_times = {detector_id: None for detector_id in edge_config.detectors.keys()}
+        # Minimum time between escalations for each detector
+        self.min_times_between_escalations = {
+            detector_id: edge_config.detectors.get(detector_id).min_time_between_escalations
+            for detector_id in edge_config.detectors.keys()
+        }
```
CoreyEWood, line 137: Tiny but you could iterate through the dict items() here to avoid the get.

app/core/edge_inference.py
CoreyEWood: Definitely agree that either getting rid of `local_inference_template` or moving the current detector config fields into it would be good, but yeah makes sense to do that in a different PR.

configs/edge-config.yaml
```diff
@@ -14,3 +14,4 @@ detectors:
       local_inference_template: "default"
       always_return_edge_prediction: false
       disable_cloud_escalation: false
+      min_time_between_escalations: None
```
CoreyEWood, line 17: I think it would be good to add some mention of this to the README. The READMEs need to get reorganized at some point so it doesn't matter too much where it goes, but having a mention of this config option in there somewhere would help make sure it makes it into the reworked version too.

CoreyEWood (approved):
Looks good! Left a few small comments.

app/api/routes/image_queries.py
f-wright: Good point! I had them on info for testing and forgot to make the change, thanks!

configs/edge-config.yaml
f-wright: Good idea! I'll add it.

PR #117, opened 2024-10-23
Fix inference model updates

CoreyEWood:
Fixes an error that arose from incorrect access of the env var attributes. Previously, the inference-model-updater would fetch a new binary but would not spawn an updated inference pod because of this error, and you would have to run cluster_setup again to use the new binary. Now, the updated model binary will actually be used once fetched.

Error that would happen:
`TypeError: 'V1EnvVar' object is not subscriptable`


tyler-romero (approved):

tyler-romero:
Thanks!

f-wright (approved):
LGTM, thanks!

PR #121, opened 2024-10-27
Setup improvements

robotrapta:


deploy/bin/setup-ee.sh
```diff
@@ -74,17 +79,17 @@ if ! ./deploy/bin/make-aws-secret.sh; then
 fi
 
 # Configmaps, secrets, and deployments
-$K delete configmap --ignore-not-found edge-config -n ${DEPLOYMENT_NAMESPACE}
```
robotrapta, line 77: The `-n namespace` flags are all redundant because it's built into `$K`.  Leaving it in is confusing.

PR #122, opened 2024-10-28
Remove IQEs

CoreyEWood:
Remove the notion of IQEs. Records of image queries generated by the edge endpoint are no longer stored locally. When returning an answer from the edge model and also escalating to the cloud, the image query ID returned by the edge endpoint will match the one in the cloud.


app/api/routes/image_queries.py
```diff
@@ -210,15 +210,3 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
         confidence_threshold=confidence_threshold,
         human_review=human_review,
     )
-
-
-@router.get("/{id}", response_model=ImageQuery)
-async def get_image_query(
```
CoreyEWood, line 216: We don't need to do any of the IQE handling here anymore, so we can just let this fall back to the cloud.

app/api/routes/image_queries.py
```diff
@@ -171,6 +170,7 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
                         patience_time=patience_time,
                         confidence_threshold=confidence_threshold,
                         human_review=human_review,
+                        image_query_id=image_query.id,
```
CoreyEWood, line 173: I think this is the only change needed for associating edge predictions with escalated queries.

tyler-romero (approved):
Nice! All looks good to me!

PR #123, opened 2024-10-31
Edge-endpoint on balena fixes

tyler-romero:
1. Fix broken ee config
2. Support bastion on x86 arch
3. Fix setup-ee.sh to work inside of containers (maybe_sudo)

### DIFF
--- a/app/core/configs.py
+++ b/app/core/configs.py
@@ -44,8 +44,8 @@ class DetectorConfig(BaseModel):
             "Requires `always_return_edge_prediction=True`."
         ),
     )
-    min_time_between_escalations: Optional[float] = Field(
-        default=None,
+    min_time_between_escalations: float = Field(
+        default=2.0,
         description=(
             "The minimum time (in seconds) to wait between cloud escalations for a given detector. "
             "Only applies when `always_return_edge_prediction=True` and `disable_cloud_escalation=False`."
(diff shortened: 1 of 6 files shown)


app/core/configs.py
```diff
@@ -44,8 +44,8 @@ class DetectorConfig(BaseModel):
             "Requires `always_return_edge_prediction=True`."
         ),
     )
-    min_time_between_escalations: Optional[float] = Field(
-        default=None,
+    min_time_between_escalations: float = Field(
+        default=2.0,
```
CoreyEWood, line 48: Should the default value be 0.0? It seems like right now the default configuration will lead to a warning being logged.

CoreyEWood (approved):
Looks good, just a question about the default min_time_between_escalations value.

app/core/configs.py
tyler-romero: Fair point about the warning. Elsewhere we default to two if this isnt provided... And two seems like a sane safety number.

CoreyEWood: The warning also isn't a huge deal, it's just warning the user that they might have misconfigured their detector. We could just let it be or change it to info. This will get solved at some point anyways when the config setup is redone.

honeytung (approved):

PR #125, opened 2024-11-01
Gpu support on top of Balena w/ k3s

tyler-romero:
Support the edge-endpoint on top of k3s on top of balena on an x86 machine + nvidia GPU.

Doing this took a bit of work plus an assist in the form of some advice from the folks at Balena.

There's a good amount of documentation in this PR about what is happening, but the high level overview is this:
1. Follow this guide to get the `server` container to have GPU access (`nvidia-smi`): https://blog.balena.io/how-to-use-nvidia-gpu-on-x86-device-balenaos/
2. Install nvidia-container-toolkit on the `server` image and update the containerd config.toml
3. Use k3s built in helm support to install the `nvidia-gpu-operator`. Configure it to NOT install nvidia drivers or nvidia-container-toolkit automatically

Additionally, to provide both GPU and continued CPU-only support, I needed to create a custom deployment script `./deploy_balena.sh`. I would LOVE suggestions for how to get rid of this or make it better. More rationale is documented on the script. I'm also worried this might conflict with glhub + edge-endpoint combined deployment.


tyler-romero:
Progress update re: gpu support for the edge-endpoint (via k3s) on balena

Whats making this hard? BalenaOS is read-only, so nvidia container toolkit can’t be installed via standard means.

BalenaOS has a guide that demonstrates how to spin up a helper pod (aka gpu)to build Nvidia drivers as kernel modules - that it then loads on the host BalenaOS. Loading kernel modules gets around the read-only host OS issue. Once these kernel modules have been loaded, nvidia-smi runs successfully on the gpu pod.

Next, we need to get nvidia-smi to run on our k3s server pod. If we install the same version of Nvidia driver here as was used on the gpu pod, then we have GPU access via nvidia-smi. This “just works”.

K3s has nice built-in support for helm via its Helm Controller, which allows just specifying Helm Charts via a Custom Resource Definition. These can be configured to install upon k3s startup, which made it easy to get the nvidia-gpu-operator installed correctly.

However, despite all of this, the k3s node still does not show GPU as a resource. Almost certainly its because `nvidia-container-toolkit` still doesnt exist on the hostOS (i did install it on the server image). Currently trying to think of a workaround for that.

deploy/balena-k3s/gpu/Dockerfile
```diff
@@ -0,0 +1,79 @@
...
+
+WORKDIR /usr/src
+
+ENV DEBIAN_FRONTEND noninteractive
+
+# Set some variables for the kernel source and NVIDIA driver
+# See https://github.com/balena-io-examples/nvidia-x86/tree/master?tab=readme-ov-file#gpu-container
+# for instructions on how to set these variables. THEY ARE IMPORTANT!
+ENV VERSION="6.0.24%2Brev1"
+# generic-aarch64
```
tyler-romero, line 11: todo

deploy/balena-k3s/README.md
```diff
@@ -1,27 +1,56 @@
...
+
+Now, we have our k3s single-node cluster built and running, but we have not started our edge deployment. To do this,
+see the [Configuration](#Configuration) section below (specifically, you will need to set the `RUN_EDGE_ENDPOINT` variable).
+
+### Deploying to an x86 device with an NVIDIA GPU
+Tested using an HP Victus Laptop with Intel Core i5 processor and NVIDIA GeForce RTX 3050 Laptop GPU.
+
+From the root of `edge-endpoint`, run:
+```bash
+./deploy_balena.sh <my-fleet> gpu
```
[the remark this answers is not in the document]
honeytung: I am also thinking about this as well. 

Right now this is probably fine because we only have either ARM CPU or x86 + CPU and we could just maintain two fleets. However, if we add Jetson into the mix (which also requires a different way to enable GPU support with a custom image for JetPack) it might become more complicated.

I am thinking maybe we can still keep the GPU container for CPU inference by modifying the launch script to not load Nvidia kernel modules and just idle instead (we can set device individual `INFERENCE_FLAVOR` in balena to control that or even modify the launch script to detect). This way we can keep all devices under the same fleet. Although this method will add more complexity so I am not sure if that is the right path we want to pursue.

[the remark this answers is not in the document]
tyler-romero: Fleets are supposed to be for one type of device in balena world are they not? I could try to see if theres a way to detect a-priori if we're running on a device with a compatible GPU

deploy/balena-k3s/README.md
```diff
@@ -1,27 +1,56 @@
...
+1. A `gpu` service that loads the compiled NVIDIA driver kernel module into the host OS, enabling GPU access for other services
+2. A `server` service running a [k3s server](https://docs.k3s.io/architecture) that acts as the k3s cluster node and installs the NVIDIA GPU operator for Kubernetes GPU support
+3. A `bastion` service that provides access to the k3s cluster via kubectl commands and contains a copy of this repo at `/app/edge-endpoint`
 
-Configure the following variables via the `<fleet>/Variables` or `<device>/Device Variables` interfaces on the BalenaCloud dashboard:
+Now, we have our k3s single-node cluster built and running, but we have not started our edge deployment. To do this,
+see the [Configuration](#Configuration) section below (specifically, you will need to set the `RUN_EDGE_ENDPOINT` variable).
+
+### Configuration
+Configure the following variables via the `<fleet>/Variables` or (less preferably) `<device>/Device Variables` interfaces on the BalenaCloud dashboard:
```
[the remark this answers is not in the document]
tyler-romero: Ah good counterexample, I'll remove that guidance

deploy/balena-k3s/README.md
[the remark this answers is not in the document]
brandon-groundlight: Keeping fleets to a given device type makes a lot of sense. The deployment process itself would be cleaner if each fleet has a single type of device, even if the number of fleets we have gets messier.

In general, I'm a fan of pushing any work that we can into the deployment and build steps. I feel like we're already suffering from slow boots and processing limits

deploy_balena.sh
```diff
@@ -0,0 +1,56 @@
...
+
+cleanup() {
+    echo "Cleaning up..."
+    rm -f docker-compose.yml
+}
+trap cleanup EXIT
+
+# Push to balena
+echo "Pushing to fleet: $FLEET_NAME with $FLAVOR configuration..."
+balena push "$FLEET_NAME"
```
honeytung, line 54: Any advice on how to update glhub so we can run edge on it?

deploy/balena-k3s/README.md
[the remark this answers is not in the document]
tomfaulhaber: The question here is "what is a 'type of device'". It's easy for that to explode, so maybe we want to spend some time thinking about that and pressure testing it.

deploy/balena-k3s/bastion/Dockerfile
```diff
@@ -63,4 +63,4 @@ RUN echo "source /app/edge-endpoint/deploy/balena-k3s/bastion/src/kube-bash.sh"
 RUN chmod +x ./edge-endpoint/deploy/bin/setup-ee.sh
 
 ENTRYPOINT []
-CMD ["/bin/sh", "-c", "./edge-endpoint/deploy/bin/setup-ee.sh && tail -f /dev/null"]
```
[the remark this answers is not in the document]
tomfaulhaber: We have some hacky code in there now that starts the containers but then just sleeps if `RUN_EDGE_ENDPOINT` isn't set.

What I'd really like to do is reconfigure all of our Balena stuff to do as much as possible in k3s and then make these decisions dynamically. This would also have the advantage of letting us use fewer fleets by having the same docker-compose setup across many different configurations.

deploy/balena-k3s/gpu/entry.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# Insert Nvidia modules
+insmod /nvidia/driver/nvidia.ko
+insmod /nvidia/driver/nvidia-modeset.ko
+insmod /nvidia/driver/nvidia-uvm.ko
+
+/usr/bin/nvidia-smi
+nvidia-modprobe
+
+sleep infinity
```
tomfaulhaber, line 18: Why not just let this container exit and set the docker-compose definition to `restart: "no"`? I don't know that makes a huge difference, but it does express intent more clearly.

It also means we could use 
```
depends_on:
  gpu:
    condition: service_completed_successfully
``` 
to have other containers wait on this getting set up, but I don't know if we need that here.

deploy/balena-k3s/server/Dockerfile
```diff
@@ -1,13 +1,45 @@
...
-RUN chmod +x /server.sh
+ARG K3S_TAG="v1.28.8+k3s1"
+
+# Use this instead of, for example, balenalib/genericx86-64-ext-ubuntu:focal, because
+# it is cross-platform and can be used on any architecture / type of device.
+FROM ubuntu:22.04
+
+ENV DEBIAN_FRONTEND noninteractive
+
+# Install Nvidia driver. Must match version in gpu container exactly
```
tomfaulhaber, line 13: Should we use a docker ARG that we set in docker-compose and pass in here? That would make it easier to change/override dynamically.

ChtGPT gives me the following docker-compose language for this:
```
  service1:
    build:
      context: ./service1
      args:
        DEPENDENCY_VERSION: "1.2.3"
```

When I push it, it says we can get fully DRY by using an `.env` file and then using environment variables for the argument.

deploy/balena-k3s/server/Dockerfile
```diff
@@ -1,13 +1,45 @@
...
+RUN \
+    apt-get update && \
+    apt install -y curl runc software-properties-common && \
+    add-apt-repository ppa:graphics-drivers/ppa && \
+    apt update && \
+    apt install -y nvidia-driver-550 \
```
tomfaulhaber, line 19: wrt the other problem we had: does this pull in `nvidia-smi` which the container toolkit depends on?

deploy/balena-k3s/server/manifests/nvidia-gpu-operator.yaml
```diff
@@ -0,0 +1,66 @@
...
+  # To disable the symlink creation, set the following envvar in ClusterPolicy:
+  #     validator:
+  #       driver:
+  #         env:
+  #         - name: DISABLE_DEV_CHAR_SYMLINK_CREATION
+  #           value: "true"
+---
+# pod to test gpu operator:
+# k logs -n gpu-operator test-cuda-vectoradd
+apiVersion: v1
```
tomfaulhaber, line 53: How much overhead does this add? Are we sharing image layers with the other GPU containers?

deploy/balena-k3s/server/server.sh
```diff
@@ -1,14 +1,35 @@
...
+    # Copy our manifests to the K3s manifests directory. Any manifests in this directory will be
+    # automatically applied to the cluster by K3s. This is how we deploy the NVIDIA GPU operator,
+    # so we only do this if we're using GPU inference.
+    # https://docs.k3s.io/installation/packaged-components#auto-deploying-manifests-addons
+    mkdir -p /var/lib/rancher/k3s/server/manifests/
+    cp /opt/k3s/manifests/* /var/lib/rancher/k3s/server/manifests/
+fi
+
+# NOTE: this script is only intended to support a single-node "cluster".
+# Multi-node clusters require using ectd as a datastore, instead of the k8s default mysql.
```
tomfaulhaber, line 30:
```suggestion
# Multi-node clusters require using etcd as a datastore, instead of the k8s default mysql.
```

tomfaulhaber (approved):
A few suggestions that you can take or leave.

Thanks for sorting this out!

deploy/balena-k3s/server/Dockerfile
tyler-romero: Yes, this is what is pulling in `nvidia-smi` here

deploy/balena-k3s/bastion/Dockerfile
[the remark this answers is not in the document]
tyler-romero: I'll remove the autostart for edge endpoint to keep the existing behavior for now

deploy/balena-k3s/server/manifests/nvidia-gpu-operator.yaml
tyler-romero: was for testing convenience mostly, will remove it

tyler-romero:
TODO: test out once again on a GPU balena device.

deploy/balena-k3s/gpu/entry.sh
tyler-romero: It turns out that Balena doesn't allow this even though its valid in docker-compose

deploy_balena.sh
tyler-romero: Will submit a PR there!

PR #127, opened 2024-11-06
Updated AWS Credentials to Fetch from Environment Variables

honeytung:
This PR contains fixes to `make-aws-secret.sh` so that the script will check if `aws configure` actually contains the credentials or not. If the CLI does not have the credentials, it will try to look for one in the environment variables.

This is useful under Balena environment where the user can just define credentials under the Balena Dashboard instead of having to manually set them from the AWS CLI.

### DIFF
--- a/deploy/bin/make-aws-secret.sh
+++ b/deploy/bin/make-aws-secret.sh
@@ -33,24 +33,16 @@ if command -v aws >/dev/null 2>&1; then
     # Try to retrieve AWS credentials from aws configure
     AWS_ACCESS_KEY_ID_CMD=$(aws configure get aws_access_key_id 2>/dev/null)
     AWS_SECRET_ACCESS_KEY_CMD=$(aws configure get aws_secret_access_key 2>/dev/null)
-
-    # Check if both keys have values
-    if [ -n "$AWS_ACCESS_KEY_ID_CMD" ] && [ -n "$AWS_SECRET_ACCESS_KEY_CMD" ]; then
-        $K create secret generic aws-credentials \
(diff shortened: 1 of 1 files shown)


CoreyEWood (approved):
If you want to avoid repeating the `$K create secret` multiple times, I think you could do this (supplied by ChatGPT and not verified to work by me):

```
if command -v aws >/dev/null 2>&1; then
    # Try to retrieve AWS credentials from aws configure
    AWS_ACCESS_KEY_ID_CMD=$(aws configure get aws_access_key_id 2>/dev/null)
    AWS_SECRET_ACCESS_KEY_CMD=$(aws configure get aws_secret_access_key 2>/dev/null)
fi

# Use configured credentials if both are available, otherwise use environment variables
AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID_CMD:-$AWS_ACCESS_KEY_ID}
AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY_CMD:-$AWS_SECRET_ACCESS_KEY}

# Create the secret with either retrieved or environment values
$K create secret generic aws-credentials \
    --from-literal=aws_access_key_id=$AWS_ACCESS_KEY_ID \
    --from-literal=aws_secret_access_key=$AWS_SECRET_ACCESS_KEY
```

But what you have now is nice and readable so I don't mind if you keep it how it is, whatever you prefer.

honeytung:
Modified the script with @CoreyEWood's suggestion and it is working fine. Will proceed with the PR with that change. 🙂

PR #129, opened 2024-11-11
Rework detector configuration

CoreyEWood:
Remove local inference templates and refactor the code accordingly. Also made various typing fixes and other small changes.

### DIFF
--- a/app/core/configs.py
+++ b/app/core/configs.py
@@ -56,6 +56,6 @@ class RootEdgeConfig(BaseModel):
 
     refresh_rate: float = Field(
         default=60.0,
-        description=("The interval (in seconds) at which the inference server checks for a new model binary update."),
+        description="The interval (in seconds) at which the inference server checks for a new model binary update.",
     )
     detectors: dict[str, DetectorConfig]
--- a/app/core/configs.py
+++ b/app/core/configs.py
@@ -1,25 +1,10 @@
 import logging
-from typing import Dict, Optional
+from typing import Optional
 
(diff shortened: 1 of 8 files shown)


app/core/configs.py
```diff
@@ -66,31 +54,8 @@ class RootEdgeConfig(BaseModel):
...
-                                    refresh_rate=120.0
-                                )
-                }
-            }
-        """
-        for detector in self.detectors.values():
-            if detector.local_inference_template not in self.local_inference_templates:
-                raise ValueError(f"Local Inference Template {detector.local_inference_template} not defined.")
-        return self
+    refresh_rate: float = Field(
```
CoreyEWood, line 57: This is now explicitly a global parameter (it actually always was, but it pretended to be detector-specific).

app/core/edge_inference.py
```diff
@@ -172,10 +174,7 @@ def detector_configured_for_local_inference(self, detector_id: str) -> bool:
         Returns:
             True if the detector is configured to run local inference, False otherwise
         """
-        if not self.inference_configs:
-            return False
-
-        return detector_id in self.inference_configs.keys() and self.inference_configs[detector_id].enabled
+        return detector_id in self.edge_config.detectors.keys() and self.edge_config.detectors[detector_id].enabled
```
CoreyEWood, line 177: Something interesting I noticed is that the `enabled` option doesn't actually prevent an inference pod from being created for a detector when False, it just prevents edge inference from occurring. I created a separate Jira ticket to address this.

PR #130, opened 2024-11-12
Setup edge deployments to 'preflight' and 'live' balena environments

roxanne-o:
I setup two fleets in balena called 'preflight' (we can use this one to run integration tests, control releases etc) and 'live' (for customer and other important edge devices). These are managed through github environments. Like in zuuul, we need an explicit approval to deploy the live fleet. Talking through with Tyler, there are a few key next steps here:
- Setting up integration tests through the preflight fleet.
- Improve unit tests
- Get live fleet into a state where we feel comfortable that it works without much manual intervention for our customers. This is where I'm most unclear about next steps.

### DIFF
--- a/.github/workflows/pipeline.yaml
+++ b/.github/workflows/pipeline.yaml
@@ -251,7 +251,10 @@ jobs:
           kubectl describe pod edge-endpoint
   
   deploy-edge-preflight-balena:
-    if: github.event_name == 'push'
+    needs: 
+      - test-general-edge-endpoint
+      - test-in-k3s
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     uses: ./.github/workflows/deploy-balena.yaml
     with:
       FLEET_NAME: preflight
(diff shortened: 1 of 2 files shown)


.github/workflows/pipeline.yaml
```diff
@@ -248,4 +248,25 @@ jobs:
         run: |
           set -ex
           kubectl rollout status deployment edge-endpoint --timeout=5m
-          kubectl describe pod edge-endpoint
\ No newline at end of file
+          kubectl describe pod edge-endpoint
+  
+  deploy-edge-preflight-balena:
+    needs:
```
brandon-groundlight, line 254: Hold up, is this going to deploy every time someone pushes to any branch?

brandon-groundlight (changes requested):
I think this will push to the fleets on any push on any branch

brandon-groundlight:
Ok, missed the check in the pre-flight fleet, but now I'm still confused why we immediately push to the live fleet after pushing to the pre-flight fleet

brandon-groundlight (approved):
Ah, we put the protections in the github settings. Mechanisms that aren't reflected in the code are always interesting

PR #131, opened 2024-11-13
Merge the k3s info into an existing kubectl config file if you're already using kubernetes on a machine

tomfaulhaber:
Also, if you're only using k3s, don't copy their config into ~/.kube since that one will be ignored.

### DIFF
--- a/deploy/bin/add-k3s-cluster-to-config.sh
+++ b/deploy/bin/add-k3s-cluster-to-config.sh
@@ -11,6 +11,14 @@ K=${KUBECTL_CMD:-kubectl}
 
 k3scfg="/etc/rancher/k3s/k3s.yaml"
 
+kubectl_is_k3s=$($K version --client | egrep "^Client Version.*+k3s")
+
+if [ -n "$kubectl_is_k3s" ]; then
+    echo "Using the kubectl supplied by k3s. Shared kubeconfig file is at ${k3scfg}"
+    echo "If you want to use a different kubeconfig, set the KUBECONFIG environment variable"
+    exit 0
+fi
+
 if [ -f $k3scfg ]; then
     # Get the cluster and user information from the k3s config file
     server=$($K config view --kubeconfig="${k3scfg}" --minify --flatten --output 'jsonpath={.clusters[?(@.name=="default")].cluster.server}')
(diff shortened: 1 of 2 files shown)


deploy/bin/add-k3s-cluster-to-config.sh
```diff
@@ -0,0 +1,43 @@
...
+
+    # Add the cluster information to the user's kubeconfig file
+    $K config set-cluster k3s --server=$server --certificate-authority=<(echo "${cert}" | base64 -d) --embed-certs=true
+    $K config set-credentials k3s \
+      --embed-certs=true \
+      --client-certificate=<(echo "${user_client_certificate}" | base64 -d) \
+      --client-key=<(echo "${user_client_key}" | base64 -d)
+
+    # create a context that points to the default namespace in the new cluster
+    $K config set-context k3s --cluster=k3s --user=k3s
```
tyler-romero, line 38: Maybe we could throw a warning or an error if there already exists a k3s context in the kubeconfig file?

deploy/bin/add-k3s-cluster-to-config.sh
```diff
@@ -0,0 +1,43 @@
...
+
+if [ -f $k3scfg ]; then
+    # Get the cluster and user information from the k3s config file
+    server=$($K config view --kubeconfig="${k3scfg}" --minify --flatten --output 'jsonpath={.clusters[?(@.name=="default")].cluster.server}')
+    cert=$($K config view --kubeconfig="${k3scfg}" --minify --flatten --output 'jsonpath={.clusters[?(@.name=="default")].cluster.certificate-authority-data}')
+
+    user_client_certificate=$(kubectl config view --kubeconfig=${k3scfg} --minify --flatten --output 'jsonpath={.users[?(@.name=="default")].user.client-certificate-data}')
+    user_client_key=$(kubectl config view --kubeconfig=${k3scfg} --minify --flatten --output 'jsonpath={.users[?(@.name=="default")].user.client-key-data}')
+
+    # Add the cluster information to the user's kubeconfig file
```
tyler-romero, line 30: Maybe we could throw a warning or an error if there already exists a k3s context in the kubeconfig file?

tyler-romero (approved):
Helpful!

deploy/bin/add-k3s-cluster-to-config.sh
tomfaulhaber: Probably shouldn't be an error and I'm not convinced it should be a warning. Generally, I think that means that we have uninstalled k3s and are reinstalling it or some variation of that. So all the cluster and user get changed and the old ones aren't valid anymore.

I suppose we could check to see if someone had a context called `k3s` that was pointing at something other than the local k3s instance, but do we think that's a real thing?

PR #132, opened 2024-11-14
Enable pulling zero-shot models and improve model updating logic

CoreyEWood:


test/edge_inference/test_model_saving.py
CoreyEWood, line 1: Git got a bit confused, but this is a renamed and modified version of `test_model_update.py`

tyler-romero (approved):
All LGTM! Nice tests too

PR #134, opened 2024-11-18
Make port 30101 work everywhere

tomfaulhaber:
This PR changes edge-endpoints default port to 30101 so that the view from outside k3s is the same as the view from direct-to-docker. I've also updated comments, readmes, etc.


tyler-romero (approved):
LGTM, thanks!

PR #138, opened 2024-11-20
Reverted buildx image back to binfmt

honeytung:
This PR reverts the `buildx` base image back to using `binfmt` because `binfmt` supports building both ARM and x86 images directly and does not require additional binaries like `alpine` does.

Note that `binfmt` does NOT support building images from ARM machines (e.g. Apple Silicon or RPis), so the build script will need to be ran on x86 machines (e.g. AWS EC2).


tyler-romero (approved):
Thank you!

PR #139, opened 2024-11-21
fix install script for k3s on new g4

brandon-groundlight:
Tried to set up an edge endpoint in my g4 and found that I needed to point Helm to use the correct kubernetes config


tomfaulhaber:
I've got a PR coming that does this in a different way.

tomfaulhaber:
I just put up [PR 140](https://github.com/groundlight/edge-endpoint/pull/140) which removes the helm part altogether by using the helmchart CRD that k3s includes by default. (Thanks to Tyler for the discovery!)

I think that fix should supersede this one.

PR #140, opened 2024-11-21
Various fixes and optimizations to the k3s deployment

tomfaulhaber:
Fixes a bunch of issues that we've seen doing installs on different kinds of systems:

1. Include poetry.lock to keep versioning consistent
2. Install curl and jq, if necessary
3. Use the helmchart CRD to install the Nvidia operator so we don't have to install and configure helm
4. Include `nvidia-utils` in the Nvidia driver install path since it's required for the operator to initialize
5. Be more patient waiting for the GPU capacity to be recognized by the node.

Reviewers could look at the specific commits for a more targeted view. This is a bit of an omnibus PR.

Tested on vanilla Ubuntu 22.04 on a G4 and on my regular G4.


brandon-groundlight (approved):
LGTM. I didn't know about helmcharts before, but after learning a bit I much prefer this over my PR

tomfaulhaber:
> LGTM. I didn't know about helmcharts before, but after learning a bit I much prefer this over my PR

@brandon-groundlight Neither did I until @tyler-romero told me about them.

Looking inside, they basically just run helm inside a container, so it's pretty simple. One nice thing is that they keep track of what's been installed and you can use `k get helmchart` to see it.

tyler-romero (approved):
LGTM!

PR #142, opened 2024-11-23
Test with k3s

roxanne-o:
Runs our preexisting docker tests in k3s. This ended up being simpler than I thought since only the live tests are used for testing a stood up service. The hardest part was getting the github actions to run. Next week I want to follow up on this by make those live tests more robust.


.github/workflows/pipeline.yaml
```diff
@@ -99,6 +99,44 @@ jobs:
         if: always()
         run: docker stop ${{ steps.start_container.outputs.container_id }}
 
+  test-with-k3s:
```
roxanne-o, line 102: Note to self: Move after build-push-edge-endpoint-multiplatform once this PR is ready to merge

test/setup_k3s_test_environment.sh
```diff
@@ -0,0 +1,49 @@
...
+
+if ! command -v k3s &> /dev/null
+then
+    echo "Error: you must have k3s setup"
+    exit 1
+fi
+export PERSISTENT_VOLUME_NAME="test-with-k3s-pv"
+export EDGE_ENDPOINT_PORT="30107"
+
+# # now we should delete the persistant volume before, in case it's in a bad state
```
tomfaulhaber, line 21:
```suggestion
# # now we should delete the persistent volume before, in case it's in a bad state
```

tomfaulhaber (approved):
LGTM

PR #145, opened 2024-11-25
Update edge submodule in glhub on merges to main

roxanne-o:
I've created a bot account that can push directly to glhub main (bypasses branch protection rule). We run this action in our protected `live` environment that me, Harry, Brandon, Corey, Tom and Tyler have access to. We approve deployments to this environment the same way we do for zuuul. Once the deployment is approved, the action will update glhub main. We've set it up this way to protect the bot account's github access token -- it lives only within the protected live environment. This action runs on merges to main of edge endpoint.


honeytung (changes requested):
The bot approach makes sense but I am not sure if we want to update all submodules. Currently it will update all submodules in the glhub repo including the `browser` submodule. I think we should limit this to only `edge-endpoint` module.

I think you can do this via
```bash
git submodule update --remote path/to/submodule
```

honeytung (approved):
LGTM! Thanks!

PR #146, opened 2024-11-25
Make the image pushing script safer

CoreyEWood:
Previously, running the `build-push-edge-endpoint-image.sh` script without any arguments would automatically push it to ECR with the `latest` tag, which is too easy to accidentally do. With these changes, you must put either `dev` or `push-latest` as the argument to make sure you're purposeful with what you're doing. 

I also updated the workflow to use this arg as well, so pushing a new `latest` image on merge should continue to work. Let me know if that part doesn't look right though.

### DIFF
--- a/deploy/bin/build-push-edge-endpoint-image.sh
+++ b/deploy/bin/build-push-edge-endpoint-image.sh
@@ -1,17 +1,16 @@
 #!/bin/bash
 
-#!/bin/bash
-
 # This script builds and pushes the edge-endpoint Docker image to ECR.
 #
 # Usage:
-#   ./build-push-edge-endpoint-image.sh [dev]
+#   ./build-push-edge-endpoint-image.sh <dev|push-latest>
 #
 # Arguments:
-#   dev: Optional. If provided, builds only for the current platform, and pushes
-#        the image to ECR *without* the :latest tag. Useful for local development.
-#        Otherwise, builds for both amd64 and arm64 architectures and pushes the
-#        image to ECR with both the :latest and :<git-tag> tags.
+#   dev:         Builds only for the current platform, and pushes the image
+#                to ECR with just the :<git-tag> tag. Useful for local development.
+#   push-latest: Builds for both amd64 and arm64 architectures and pushes the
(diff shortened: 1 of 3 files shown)


deploy/bin/build-push-edge-endpoint-image.sh
```diff
@@ -37,8 +50,10 @@ aws ecr get-login-password --region us-west-2 | docker login \
...
                   --password-stdin  ${ECR_URL}
 
-# Check if the first argument is "dev". If it is, only build the image for the current platform
+# Check if the first argument is "dev". If it is, only build the image for the current platform and don't push the
+# :latest tag.
 if [ "$1" == "dev" ]; then
+  echo "Received argument 'dev'. Building for current platform only. Latest tag will not be pushed."
   docker build --tag ${EDGE_ENDPOINT_IMAGE} ../..
   docker tag ${EDGE_ENDPOINT_IMAGE}:latest ${ECR_URL}/${EDGE_ENDPOINT_IMAGE}:${TAG}
   docker push ${ECR_URL}/${EDGE_ENDPOINT_IMAGE}:${TAG}
```
honeytung, line 59: Since we are proposing using `dev` tag for safer testing, I think we should also make sure that we can build the image for both ARM and x86 so that people testing it out with ARM devices can also run the `dev` images.

We could just move this portion after buildx command use it:
```
# Build image for amd64 and arm64
docker buildx build \
  --platform linux/arm64,linux/amd64 \
  --tag ${ECR_URL}/${EDGE_ENDPOINT_IMAGE}:${TAG} \
  ../.. --push
```

honeytung (changes requested):
This looks good, although I think we should change this script or the inference image build script so that their commands are similar just to be consistent but that is not necessarily for now. Left a comment about support both ARM and x86 images for testing. 🙂

tomfaulhaber:
I am closing this PR because it's solved in [COM-2234: Use Helm to deploy Edge Endpoint by tomfaulhaber · Pull Request #184 · groundlight/edge-endpoint](https://github.com/groundlight/edge-endpoint/pull/184)](https://github.com/groundlight/edge-endpoint/pull/184)

PR #147, opened 2024-12-05
Increase frequency of confidence threshold updates

CoreyEWood:
Previously, detector metadata (including confidence threshold) would be cached for an hour, meaning updates to the confidence threshold via the dashboard wouldn't be incorporated for a while. This was liable to create confusing behavior, where a user adjusts the confidence threshold but sees unexpected behavior on the edge for a while afterward. The updated value would only be received once the cached value expired or upon a re-deploy.

Now, cached detector metadata can only stick around for a maximum of 10 minutes, and otherwise will be re-computed once an image query is sent in. Additionally, a background task runs after each image query request is processed which checks how old the cached value for a detector's metadata is. If it's older than 30 seconds, it will be refreshed (still in the background, meaning it doesn't slow down query processing times - other than occupying the worker for slightly longer). 

Let me know if this seems like it's over-complicating things too much. I tried to avoid defining a new cache class but wasn't able to figure out a better way to do this. Definitely open to ideas though!


app/core/app_state.py
```diff
@@ -12,13 +12,14 @@
...
-from .utils import safe_call_sdk
+from .utils import TimestampedTTLCache, safe_call_sdk
 
 logger = logging.getLogger(__name__)
 
 MAX_SDK_INSTANCES_CACHE_SIZE = 1000
 MAX_DETECTOR_IDS_TTL_CACHE_SIZE = 1000
-TTL_TIME = 3600  # 1 hour
+TTL_TIME = 600  # 10 minutes
+STALE_METADATA_THRESHOLD = 30  # 30 seconds
```
tyler-romero, line 22: style nit: it can be nice to define time values, like `TEN_MINUTES = 60 * 10` and then to use these constants directly. Like `if cached_value_age > THIRTY_SECONDS:` below.
Separately, its a good idea to include units in variable names. So instead of `STALE_METADATA_THRESHOLD`, `STALE_METADATA_THRESHOLD_SEC`

app/core/utils.py
```diff
@@ -174,6 +175,27 @@ def pil_image_to_bytes(img: Image.Image, format: str = "JPEG") -> bytes:
         return buffer.getvalue()
 
 
+class TimestampedTTLCache(cachetools.TTLCache):
+    """TTLCache subclass that tracks when items were added to the cache."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.__timestamps = {}  # Store timestamps for each key
```
tyler-romero, line 183: why the double-underscore ("name mangling") here? https://realpython.com/python-double-underscore/#double-leading-underscore-in-classes-pythons-name-mangling

CoreyEWood: I used Cursor to help auto-generate this and the classes that this inherits from makes use of the double underscores, so I left it in to keep it consistent with those cachetools classes. But reading this it seems like it doesn't provide any value for the timestamps attribute, so I'll remove it.

tyler-romero (approved):
LGTM, thanks!

honeytung (approved):
LGTM! Nice QoL changes!

PR #149, opened 2024-12-09
Integration tests

roxanne-o:
These tests expand on the basic integration we had for k3s. They are a bit unusual as they weave python and bash together. We test that we can setup the inference pods, submit to edge and get a low confidence answer at first, submit to edge and escalate to cloud to train the model, pull the updated model back down to the cloud into a new inference pod, and now make a confident edge prediction.


test/integration/dog.jpg
tomfaulhaber, line 1: That's a good dog! 14/10!

test/integration/integration_test.py
```diff
@@ -0,0 +1,144 @@
...
+
+import argparse
+import random
+import time
+
+from groundlight import Groundlight, GroundlightClientError
+from model import Detector
+
+NUM_IQS_TO_IMPROVE_MODEL = 10
+ACCETABLE_TRAINED_CONFIDENCE = 0.8
```
tomfaulhaber, line 18:
```suggestion
ACCEPTABLE_TRAINED_CONFIDENCE = 0.8
```
And wherever this value is used

tomfaulhaber (approved):
LGTM with one spelling fix

test/integration/integration_test.py
```diff
@@ -0,0 +1,144 @@
...
+def main():
+    parser = argparse.ArgumentParser(
+        description="Submit a dog and cat image to k3s Groundlight edge-endpoint for integration tests"
+    )
+    parser.add_argument(
+        "-m",
+        "--mode",
+        type=str,
+        choices=["create_detector", "initial", "improve_model", "final"],
+        help="Mode of operation: 'initial', 'many', or 'final'",
```
CoreyEWood, line 43: Should this have "create_detector", "initial", "improve_model", and "final" as the options? Or is this saying something different?

test/integration/integration_test.py
```diff
@@ -0,0 +1,144 @@
...
+    elif args.mode == "improve_model":
+        improve_model(detector)
+    elif args.mode == "final":
+        submit_final(detector)
+
+
+def create_cat_detector() -> str:
+    """Create the intial cat detector that we use for the integration tests. We create
+    a new one each time."""
+    random_number = random.randint(0, 9999)
```
CoreyEWood, line 70: Is it worth increasing the range here just to make it even more unlikely that we get a collision? You could also generate a ksuid like we do [here](https://github.com/groundlight/edge-endpoint/blob/2db00f7568469ac0d09b357ae349f2a0c6ee65b0/app/core/utils.py#L157) to ensure there's no issues.

test/integration/integration_test.py
```diff
@@ -0,0 +1,144 @@
...
+    # 0.5 threshold to ensure we get a edge answer
+    iq_yes = _submit_cat(detector, confidence_threshold=0.5)
+    iq_no = _submit_dog(detector, confidence_threshold=0.5)
+    end_time = time.time()
+    print(f"Time taken to get low confidence response from edge: {end_time - start_time} seconds")
+
+    # a bit dependent on the current default model,
+    # but that one always defaults to 0.5 confidence at first.
+    assert iq_yes.result.confidence == 0.5
+    assert iq_no.result.confidence == 0.5
```
CoreyEWood, line 89: At some point we're planning to make the default edge binary pipeline be our normal default binary pipeline, which does make actual zeroshot predictions (which are still close to 0.5, but not exactly 0.5). Maybe this should check that the confidence is in a slightly wider range? I'm worried we won't remember to update this when we change the default edge pipeline.

test/integration/integration_test.py
```diff
@@ -0,0 +1,144 @@
...
+## - Submit the final dog/cat image query to the edge, expect high confidence
+
+import argparse
+import random
+import time
+
+from groundlight import Groundlight, GroundlightClientError
+from model import Detector
+
+NUM_IQS_TO_IMPROVE_MODEL = 10
```
CoreyEWood, line 17: This name is slightly misleading because I think double this amount of IQs actually get submitted? Maybe it should be renamed to something like`NUM_IQS_PER_CLASS_TO_IMPROVE_MODEL`, though I don't think it's too important either way.

test/integration/run_tests.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# Ensure our most recent pod is brand new.
+most_recent_pod=$(kubectl get pods -n $DEPLOYMENT_NAMESPACE -l app=inference-server -o jsonpath='{.items[-1].metadata.name}')
+current_time=$(date +%s)
+pod_creation_time=$(kubectl get pod $most_recent_pod -n $DEPLOYMENT_NAMESPACE -o jsonpath='{.metadata.creationTimestamp}')
+pod_creation_time_seconds=$(date -d "$pod_creation_time" +%s)
+time_difference=$((current_time - pod_creation_time_seconds))
+
+
+# Check if the pod was created within 1.1 times the refresh rate
```
CoreyEWood, line 38: Should say 3 times the refresh rate here I think.

CoreyEWood (approved):
Left some various comments about small things, but overall this is great! Very excited to have this.

PR #150, opened 2024-12-09
Use groundlight 0.21.0 (and also pillow 11.0.0)

CoreyEWood:
Bumping groundlight SDK version. Also requires us to use pillow ^11.0.0. According to [the docs](https://pillow.readthedocs.io/en/stable/releasenotes/11.0.0.html), I don't think that should introduce any breaking changes for us.


brandon-groundlight (approved):
LGTM

honeytung (approved):
Was scared at first at how many lines it changed but realized it is just poetry. Otherwise LGTM!

PR #152, opened 2024-12-19
Include metadata identifying edge results

CoreyEWood:
When an IQ is created and returned from the edge, it will now have `is_from_edge: True` in the metadata so it can be identified as an edge answer.


honeytung (approved):
LGTM thanks!

PR #153, opened 2024-12-20
Load testing README updates and small improvements

CoreyEWood:
Updates the README to be accurate to the current configuration options. Adds a config variable to check the returned number of objects when using a count detector. Removes an unnecessary print statement.


honeytung (approved):
LGTM! 🙂

PR #156, opened 2025-01-02
Audit confident edge predictions

CoreyEWood:
When the edge makes a confident prediction there is now a chance (defined by the `confident_audit_rate` parameter) that the query will be sent to the cloud to be audited after the edge answer is returned. This allows us to better evaluate the accuracy of the edge pipeline. 

The escalated image query is submitted with metadata `is_edge_audit = True`, which will cause it to be audited in the cloud regardless of the cloud ML result.

This change is fully backwards-compatible because the new parameter will assume the default value if not present in the config. I tested this by switching to the new edge-endpoint image on an already-running edge endpoint and verified that it still worked correctly.

The current default rate is 1e-5, meaning there is a 0.001% chance to audit a confident query.

There is currently no way to disable auditing (other than disabling cloud escalations entirely), but I could add that if we think that might be desired. 

The `post_image_query` method is getting huge and confusing so I think we should move more of the code into helper functions, but I plan to do that in a separate PR.


configs/edge-config.yaml
```diff
@@ -1,7 +1,8 @@
 # For configuring detectors on the edge endpoint. See CONFIGURING-DETECTORS.md for more information.
 
 global_config: # These settings affect the overall behavior of the edge endpoint.
-  refresh_rate: 60 # How often to attempt to fetch updated ML models (in seconds). If not set, defaults to 60.
+  refresh_rate: 60 # How often to attempt to fetch updated ML models (in seconds). Defaults to 60.
+  confident_audit_rate: 0.001 # Probability that a confident prediction will be sent to cloud for auditing. Defaults to 0.001.
```
paulina-positronix, line 5: So, what's the standard FPS rate on an edge detector (binary)? If it's 5, then that's a human-required escalation for audit every 3.5 minutes or so, I think. Seems like too much.

app/api/routes/image_queries.py
```diff
@@ -164,9 +165,34 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
...
+                        safe_call_sdk,
+                        gl.submit_image_query,
+                        detector=detector_id,
+                        image=image_bytes,
+                        wait=0,
+                        patience_time=patience_time,
+                        confidence_threshold=confidence_threshold,
+                        human_review="ALWAYS",  # Require human review for audited queries so we can evaluate accuracy
+                        want_async=True,
+                        metadata={"is_edge_audit": True},  # Provide metadata to identify edge audits in the cloud
```
paulina-positronix, line 187: On the cloud audits get a lower review priority than escalations. How if at all is that handled here? When combined with the audit rate / sheer number of audited images, this is a potential area of concern.

paulina-positronix (commented):
I have a couple of questions/concerns around human review, audit rate and review priority

configs/edge-config.yaml
CoreyEWood: Definitely a good point! It's hard to say what the standard FPS rate is since there's such a wide range of use cases, which makes it hard to pick a rate that works for all cases. Something approximating video processing (or a setup with multiple cameras going to a single detector) could definitely be 5 FPS or more. For high FPS use cases the rate could be configured to be lower (though I don't think we could expect a customer to do that without direct guidance from us). A concern with lowering it too much would be that it becomes much less useful for lower FPS cases.

tyler-romero: We could instead audit on a timer, so our audit frequency is independent of FPS, but thats no longer random sampling. What do you think Paulina?

app/api/routes/image_queries.py
tyler-romero: Yeah we do want to treat edge audits in the same way as cloud audits, which will likely take some backend work.

CoreyEWood: With the current code, the 'audit' escalations to the cloud will be treated as normal escalations, so they won't have the reduced priority that cloud audits do. I tried to find a way for these escalations to be counted as audits, but I'm pretty sure there's no way to do it without making backend changes (and potentially SDK changes as well). Given that, I thought it could be worth introducing edge audits via this simple approach first, but we could go with the more complicated approach instead if that seems worthwhile.

configs/edge-config.yaml
paulina-positronix: I think that's fine

CoreyEWood: I think the timer would some additional complexity with our multiple worker setup because each worker would have their own timer unless we make the timer external to the processes somehow - which is possible, just maybe more complicated than we'd want.

app/api/routes/image_queries.py
tyler-romero: > reduced priority that cloud audits do
Sounds like they also wont be sure to get cloud labels though. I dont think thats a showstopper for now but I would add some comments to make that loud and obvious to people looking through this code.

app/api/routes/image_queries.py
```diff
@@ -168,9 +169,31 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
                 text=results["text"],
             )
 
-            # Escalate after returning edge prediction if escalation is enabled and we have low confidence
```
robotrapta, line 171: Not needed for this PR, but we need to factor this method.  One function is hundreds of lines of code and the entire file.

app/core/configs.py
```diff
@@ -11,6 +11,10 @@ class GlobalConfig(BaseModel):
         default=60.0,
         description="The interval (in seconds) at which the inference server checks for a new model binary update.",
     )
+    confident_audit_rate: float = Field(
+        default=0.001,
```
robotrapta, line 15:
```suggestion
        default=1e-5,
```
I think we need to set this quite a bit lower.  If somebody is running 10fps constantly we're seeing 1M iq/day, so this would escalate 1,000 images per day.  I think escalating 1 image/day if we're doing 1fps is pretty reasonable, which is about 1e-5.

app/core/configs.py
```diff
@@ -84,6 +88,7 @@ def validate_inference_configs(self):
             {
                 'global_config': {
                     'refresh_rate': 60.0,
+                    'confident_audit_rate': 0.001,
```
robotrapta, line 91:
```suggestion
                    'confident_audit_rate': 1e-5,
```

configs/edge-config.yaml
robotrapta: Timer is definitely better for most cases.  If we turn this on will we have backwards compatibility issues when we add timer later?  I think so, which is maybe a good reason to fix it before we merge this.

Without timer, I think 1e-5 fraction is a decent place to start.  1fps translates to 1 audit/day.

robotrapta (commented):
I think we shouldn't merge this until we change to a timer-based approach, because of the backwards compatibility issues it would introduce.  Unless we support both kinds of config, which seems unnecessarily complicated.

configs/edge-config.yaml
paulina-positronix: 1 audit/hour would be my preference

app/api/routes/image_queries.py
CoreyEWood: Totally agree, this has been something I've wanted to do for a while. We have pretty poor testing on this function so I wanted to add more tests before making changes to make sure no bugs get inadvertently introduced, which I got pretty far on before getting pulled into other things. So this should be possible pretty soon.

tyler-romero (commented):
Commenting just to clear up my PR review queue. Click re-request review when ready!

configs/edge-config.yaml
```diff
@@ -1,7 +1,8 @@
 # For configuring detectors on the edge endpoint. See CONFIGURING-DETECTORS.md for more information.
 
 global_config: # These settings affect the overall behavior of the edge endpoint.
-  refresh_rate: 60 # How often to attempt to fetch updated ML models (in seconds). If not set, defaults to 60.
+  refresh_rate: 60 # How often to attempt to fetch updated ML models (in seconds). Defaults to 60.
+  confident_audit_rate: 0.00001 # Probability that a confident prediction will be sent to cloud for auditing. Defaults to 1e-5 = a 0.001% chance.
```
CoreyEWood, line 5: I made the default 1e-5 per Leo's suggestion, but can change that if others think it should be higher.

tyler-romero: I think this makes sense. 1/100000 IQs and those IQs should be coming pretty fast.

app/api/routes/image_queries.py
```diff
@@ -168,9 +169,34 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
...
-            if not disable_cloud_escalation and not is_confident_enough:
-                # Only escalate if we haven't escalated on this detector too recently
+            # Skip cloud operations if escalation is disabled
+            if disable_cloud_escalation:
+                return image_query
+
+            if is_confident_enough:  # Audit confident edge predictions at the specified rate
+                if random.random() < app_state.edge_config.global_config.confident_audit_rate:
```
tyler-romero, line 177: Makes sense. Possible to add a test for this codepath? Eg by setting confident_audit_rate to 100% and then ensuring this works as expected?

tyler-romero (approved):
LGTM, would like to see a test for this behavior if possible.

tyler-romero (approved):

PR #157, opened 2025-01-08
CICD: Runs a full GPU install on an EC2 instance

robotrapta:
Pretty big (but useful!) change.  Adds a GHA workflow step that in a fully automatic way:

- Creates a new g4 EC2 instance
- Installs EE on it.  (Installs K3s, and installs our YAMLs.)
- Checks that the EE install script runs successfully
- Checks that the k8 deployments come up
- Checks that the SDK can do minimal things through the EE (whoami, list-detectors)
- (Does not tears down the EC2 infra - there's a sweeper which will run every 30 minutes and do that async so the pipeline doesn't have to wait for it, because it takes ~7 minutes to turn off a G4 instance.)

It uses pulumi for infra, and relies on pulumi infra defined in the GL_Public AWS account.

It does NOT yet:
- Check that any inference models work.

(Note: there is replication of #169 which improves the workflow YAML and validation thereof directly.)
(Note: this relies on resources defined in the `gl_public` account defined in the internal `ci-infra` repo.)

### DIFF
--- a/.github/workflows/pipeline.yaml
+++ b/.github/workflows/pipeline.yaml
@@ -28,7 +28,7 @@ jobs:
 
       - name: Set up python
         id: setup_python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ env.PYTHON_VERSION }}
 
@@ -179,7 +179,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Set up python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ env.PYTHON_VERSION }}
 
@@ -290,7 +290,7 @@ jobs:
           aws sts get-caller-identity
 
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ env.PYTHON_VERSION }}
 


.github/.yamllint.yaml
robotrapta, line 1: This (and a couple other YAML changes) are isolated in #168

.github/workflows/pipeline.yaml
```diff
@@ -1,6 +1,9 @@
 name: cicd
 on:
-  push:
```
robotrapta, line 3: "push" means any push, even if not in a PR.  I don't think we want that.

deploy/bin/install-k3s-nvidia.sh
```diff
@@ -60,14 +60,15 @@ echo "NVIDIA GPU Operator installation completed."
 # Verify that we actually added GPU capacity to the node
 capacity=0
 elapsed=0
-timeout=120
+timeout_min=5
```
robotrapta, line 63: Sometimes GPUs are slow...  (news flash I know)

robotrapta:
Okay I think this is actually working again and really ready for review!  I'm gonna deliberately break the k8 YAML and make sure the test fails (it wouldn't have before) and then restore it.

robotrapta:
:)

```
Error from server (BadRequest): error when creating "deploy/k3s/edge_deployment/edge_deployment.yaml.tmp": Deployment in version "v1" cannot be handled as a Deployment: strict decoding error: unknown field "shrek"
```

cicd/pulumi/.envrc
```diff
@@ -0,0 +1,3 @@
+echo "This is a uv project.  Remember to 'uv run ...' everything"
```
tyler-romero, line 1: Oooohhh

cicd/pulumi/uv.lock
tyler-romero, line 1: Is this a worry b/c dependabot wont check uv.lock files yet?

tyler-romero (approved):
I mean it looks good to me!

cicd/pulumi/uv.lock
robotrapta: It's not ideal.  But this dependency closure is pretty tiny - only a handful of packages in here.  And it looks like dependabot is going to add uv this quarter.

PR #158, opened 2025-01-09
Adding liveness probes so k8 will restart failed pods.

robotrapta:
We've seen inference worker pods fail and not get automatically restarted.  This adds liveness probes so it will.


tyler-romero (approved):
LGTM

PR #159, opened 2025-01-12
Automatically refresh the ECR login credentials

robotrapta:
Cron-job that every 2 hours refreshes the ECR login token.

I put all the needed functionality into the `edge-endpoint` container, which adds a few tools that weren't there before, so it's bigger.  I also gave that container authz to create/destroy secrets.  We could isolate this stuff more, but it would be more complex.  The auth stuff is limited to the namespace. 

Tasks:
- [x] Get the refresh script into Dockerfile
- [x] Fully test


deploy/k3s/refresh_creds.yaml
```diff
@@ -0,0 +1,37 @@
...
+      backoffLimit: 3
+      template:
+        metadata:
+          labels:
+            app: refresh_creds
+        spec:
+          serviceAccountName: edge-endpoint-service-account
+          containers:
+          - name: ecr
+            image: 767397850842.dkr.ecr.us-west-2.amazonaws.com/edge-endpoint:${IMAGE_TAG}
```
tomfaulhaber, line 19: Do you want to parameterize account and region here like you did above? Seems like we should be consistent on that one.

deploy/bin/setup-ee.sh
```diff
@@ -142,6 +142,14 @@ $K get service -o custom-columns=":metadata.name" --no-headers=true | \
     grep "inference-service" | \
     xargs -I {} $K delete service {}
 
+
+# Define a function to envsubst and apply a yaml file
+apply_yaml() {
```
tomfaulhaber, line 147: Nice!!

tomfaulhaber (approved):
LGTM - just one nit about passing through the account and region so it's consistently handled.

robotrapta:
Corey found an issue in manual testing, so I'm going to wait to merge this until A) I have [end-to-end tests automated](https://github.com/groundlight/edge-endpoint/pull/157) (preferable) or B) I track down and fix the problem.

PR #163, opened 2025-01-14
Fix persistent volume conflict check

CoreyEWood:
Now that we don't assign the local persistent volume a storage class, the conflict check was failing every time I re-ran `setup-ee.sh` after the first time because it was still expecting the PV to have `local-sc` storage class.

Also fixes a small typo I noticed.


tomfaulhaber (approved):
Thanks. Good catch.

PR #169, opened 2025-01-19
Deeper validation of workflow yamls

robotrapta:
Extension to #168 - this installs a semantic checker for workflow yamls. (Not just YAML linter.)

It runs a strict test on for the YAMLs for mistakes, and then an informational test with all the linting warnings about deprecations and such.

### DIFF
--- a/.github/workflows/pipeline.yaml
+++ b/.github/workflows/pipeline.yaml
@@ -24,6 +24,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Set up python
+        id: setup_python
         uses: actions/setup-python@v4
         with:
           python-version: ${{ env.PYTHON_VERSION }}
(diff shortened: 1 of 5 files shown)


.github/workflows/pipeline.yaml
```diff
@@ -1,6 +1,9 @@
 name: cicd
 on:
-  push:
+  pull_request:
+    branches:
+      - main
+    types: [opened, synchronize, reopened]
```
tyler-romero, line 6: Is this excluding something in particular?

tyler-romero (approved):

.github/workflows/pipeline.yaml
robotrapta: Not sure TBH.

PR #170, opened 2025-01-19
E2e tests on k8, minimal SDK functionality

robotrapta:
An extension of #157 (and I might merge it into that one) which adds the next step of useful tests:

- Checks that the k8 deployments come up
- Checks that the SDK can do minimal things through the EE (whoami, list-detectors)


robotrapta:
Merged into 157

PR #172, opened 2025-01-19
E2e test of refresh ecr

robotrapta:
Using the e2e-testing to see if #159 is working.


robotrapta:
This was an experiment to try to repro a manually-discovered problem.  This test didn't discover it.

PR #174, opened 2025-02-04
Turning on the CICD pipeline on merges to main!

robotrapta:
Also, making it so the e2e test machine stays awake if it fails its tests.


tomfaulhaber (approved):
LGTM!

PR #175, opened 2025-02-05
Save OODD models to repository

f-wright:
First edge-endpoint PR to get OODD running on the edge! This PR makes use of the new information provided through the fetch-model-urls endpoint, and saves OODD models to the edge. It sets up a new model repository format, where the primary edge model versions are saved under `<detector_id>/primary` within the model repository, instead of just `<detector_id>`. The new OODD models will be saved under `<detector_id>/oodd`. 

To switch existing edge-endpoints over to the new model repository format, the edge will pull the latest model for initial set up, and clear out the old model versions in the repository. This will happen even if the cloud model is the same as the current edge model. 

In this PR, nothing is done to use the downloaded OODD model. Future PRs will spin up a separate inference pod for it, and combine the primary and OODD predictions into a single prediction.


test/edge_inference/test_model_saving.py
```diff
@@ -63,43 +81,173 @@ def test_save_model_with_no_binary_to_repository():
...
+        assert not os.path.exists(os.path.join(oodd_model_dir, "2"))
+
+        # and both primary and oodd should have versions 3 and 4
+        assert os.path.exists(os.path.join(primary_model_dir, "3"))
+        assert os.path.exists(os.path.join(oodd_model_dir, "3"))
+        assert os.path.exists(os.path.join(primary_model_dir, "4"))
+        assert os.path.exists(os.path.join(oodd_model_dir, "4"))
+
+
+def test_switch_to_new_model_repository_format():
```
tyler-romero, line 194: nice! was looking for details on this

tyler-romero (commented):
Overall looks really nice - before approving I wanted to check if youve tried this on an actual edge-endpoint to see if it keeps running correctly after switching to the new model repo format?

f-wright:
> Overall looks really nice - before approving I wanted to check if youve tried this on an actual edge-endpoint to see if it keeps running correctly after switching to the new model repo format?

Yep! I tested this out on both beebop and a g4. The old models persist for a little bit after new models are pulled for the new model repository format, since `delete_old_model_versions` isn't called when the first deployment is created for a newly set up edge endpoint. Once it goes through the update deployment code path, the old models are deleted. There aren't any interruptions to running the new edge endpoint.

app/core/edge_inference.py
```diff
@@ -221,47 +221,44 @@ def run_inference(self, detector_id: str, image_bytes: bytes, content_type: str)
         logger.info(f"Recent-average FPS for {detector_id=}: {fps:.2f}")
         return output_dict
 
-    def update_model(self, detector_id: str) -> bool:
+    def update_models_if_available(self, detector_id: str) -> bool:
```
CoreyEWood, line 224: I like how you reorganized this function! Definitely makes it easier to follow.

CoreyEWood (approved):
Looks great! I really like how you organized the code, and the tests are excellent. The additional logging is also great.

PR #177, opened 2025-02-07
Off by one error in installation script

f-wright:
This PR fixes a small bug I encountered when setting up a new edge-endpoint on a g4.


robotrapta (approved):
Thanks!

PR #178, opened 2025-02-08
Add a script to update the system when the IP address changes

tomfaulhaber:
To try this on a system with a changing IP address:

```
# Get the script
cd .../edge-endpoint
git pull
git checkout origin/ip-change-script

# Run the script
deploy/bin/ip-change.sh
```

I'll add some verbiage to the README on Monday.

### DIFF
--- a/deploy/bin/ip-changed.sh
+++ b/deploy/bin/ip-changed.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+
+# If the IP is changed, we'll need to restart some parts of the Kubernetes stuff
+# Basically the K8s "node" object has the IP address in it and that doesn't change
+# when the IP address changes. So we delete the node and restart the k3s service which
+# will create a new node with the right address.
+
+# Right now, this is pretty simple and basically recreates the whole edge endpoint.
+# I think that we could get a little more fine-grained with this and figure out how
+# to just recreate the persistent volumes cleanly, but this is good enough for now.
+
+set -e # exit on error
+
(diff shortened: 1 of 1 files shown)


tyler-romero (approved):
Nice and easy to understand, LGTM

deploy/bin/ip-changed.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+# If the IP is changed, we'll need to restart some parts of the Kubernetes stuff
+# Basically the K8s "node" object has the IP address in it and that doesn't change
+# when the IP address changes. So we delete the node and restart the k3s service which
+# will create a new node with the right address.
+
+# Right now, this is pretty simple and basically recreates the whole edge endpoint.
```
brandon-groundlight, line 8: We wipe the model binaries too then and have to redownload. But I definitely agree, this is good now

brandon-groundlight (approved):
Makes sense, thanks for solving yet another mystery

deploy/bin/ip-changed.sh
brandon-groundlight: Tom pointed out this isn't true - the data will still live on the host and when we recreate the pvs they'll find the existing downloads again!

tomfaulhaber: The data isn't deleted because the PVs point at a host directory whose contents are maintained through this cycle of deleting the PVC/PV (which are just k8s constructs in this case) and restoring them.

So, we shouldn't need to redownload the model binaries.

PR #179, opened 2025-02-10
Perform OODD inference on the edge

f-wright:
This is the 2nd PR to get OODD inference running on the edge. The [first PR](https://github.com/groundlight/edge-endpoint/pull/175) downloaded the models to the edge and saved them in the model repository. This PR takes those saved OODD models for deployed detectors, spins up an inference pod for them alongside the primary inference pod, and runs OODD inference in parallel with primary inference.

In this PR, the OODD inference result isn't used yet. A third PR will take the inference results from the primary and OODD models and combine them into a final edge prediction.


app/core/edge_inference.py
```diff
@@ -33,18 +35,20 @@ def is_edge_inference_ready(inference_client_url: str) -> bool:
...
-    try:
-        response = requests.post(inference_url, data=image_bytes, headers=headers)
-        if response.status_code != status.HTTP_200_OK:
-            logger.error(f"Inference server returned an error: {response.status_code} - {response.text}")
-            raise RuntimeError(f"Inference server error: {response.status_code} - {response.text}")
-        return response.json()
-    except requests.exceptions.RequestException as e:
-        logger.error(f"Failed to connect to {inference_url}: {e}")
-        raise RuntimeError("Failed to submit image for inference") from e
+    async with httpx.AsyncClient() as client:
```
tyler-romero, line 41: Just confirming - does httpx raise the same `requests.exceptions.RequestException` exception?

app/core/edge_inference.py
```diff
@@ -183,16 +194,17 @@ def inference_is_available(self, detector_id: str) -> bool:
...
             inference_client_url = self.inference_client_urls[detector_id]
+            oodd_inference_client_url = self.oodd_inference_client_urls[detector_id]
         except KeyError:
-            logger.debug(f"Failed to look up inference client for {detector_id}")
+            logger.info(f"Failed to look up inference clients for {detector_id}")
             return False
 
-        if not is_edge_inference_ready(inference_client_url):
-            logger.debug("Edge inference server is not ready")
+        if not (is_edge_inference_ready(inference_client_url) and is_edge_inference_ready(oodd_inference_client_url)):
```
tyler-romero, line 202: nit: imo `not A or not B` is a bit easier to reason about than `not (A and B)`

In general one goal of writing code should be to reduce cognitive overhead, and so another strategy is to pull out logic into a boolean variable, like:

```
inference_servers_are_ready = is_edge_inference_ready(inference_client_url) and is_edge_inference_ready(oodd_inference_client_url)
if not inference_servers_are_ready:
	...
```

tyler-romero (approved):
Looks great! Thanks!

app/core/edge_inference.py
f-wright: Good catch, thank you. No, it doesn't, updating it to the httpx error.

app/core/edge_inference.py
```diff
@@ -183,16 +194,20 @@ def inference_is_available(self, detector_id: str) -> bool:
         """
         try:
             inference_client_url = self.inference_client_urls[detector_id]
+            oodd_inference_client_url = self.oodd_inference_client_urls[detector_id]
         except KeyError:
-            logger.debug(f"Failed to look up inference client for {detector_id}")
+            logger.info(f"Failed to look up inference clients for {detector_id}")
```
CoreyEWood, line 199: Looking at this, I wonder if we should change the wording of this log - failed feels like it implies that it should have been able to find the inference clients, but if this is the first request for the detector then it's expected that it wouldn't find them. Maybe it should say something like "Did not find inference clients for {detector_id}" instead.

app/core/models.py
```diff
@@ -20,7 +20,14 @@ class InferenceDeployment(Base):
...
     __tablename__ = "inference_deployments"
-    detector_id = Column(String(44), primary_key=True, unique=True, nullable=False, comment="Detector ID")
+    model_name = Column(
+        String(100),
+        primary_key=True,
+        unique=True,
+        nullable=False,
+        comment="Name of the model, detector_id + `/primary` or `/oodd`",
+    )
+    detector_id = Column(String(44), primary_key=False, nullable=False, comment="Detector ID")
```
CoreyEWood, line 30: Is this column only here for backwards compatibility? If so, could be good to leave a comment to make that clear.

app/model_updater/update_models.py
```diff
@@ -44,39 +49,59 @@ def _check_new_models_and_inference_deployments(
...
+
+    if oodd_deployment is None:
+        logger.info(f"Creating a new oodd inference deployment for {detector_id}")
+        deployment_manager.create_inference_deployment(detector_id=detector_id, is_oodd=True)
 
     if new_model:
         # Update inference deployment and rollout a new pod
         logger.info(f"Updating inference deployment for {detector_id}")
         deployment_manager.update_inference_deployment(detector_id=detector_id)
+        deployment_manager.update_inference_deployment(detector_id=detector_id, is_oodd=True)
```
CoreyEWood, line 69: If only one of the edge and oodd models was updated, will a new inference pod be created for both? Or only for the one that had an update?

deploy/README.md
```diff
@@ -61,12 +61,13 @@ NAME                                    READY   STATUS    RESTARTS   AGE
 edge-endpoint-594d645588-5mf28          2/2     Running   0          4s
 ```
 
-If you configured detectors in the [edge config file](/configs/edge-config.yaml), you should also see a pod for each of them, e.g.:
+If you configured detectors in the [edge config file](/configs/edge-config.yaml), you should also see 2 pods for each of them (one for primary inference and one for OODD), e.g.:
```
CoreyEWood, line 64: Should we explain in the README somewhere what OODD means and why there's a pod for it?

CoreyEWood (commented):
This looks really good! Left a few comments about small clarifications.

Just as a safety check, are we okay with all existing edge deployments spawning OODD pods once this is merged? If there are any concerns about that negatively affecting anything we could consider gating this behind an env var or something like that. Initially though I don't think I can foresee any issues from it (other than maybe users being a little confused).

app/core/models.py
f-wright: No, I mostly left it in because I just thought it was useful. Seems like we could want to manage deployments by detector in the future, and it felt unnecessary to take it out.

PR #181, opened 2025-02-13
Updating documentation in CONFIGURING-DETECTORS.md

timmarkhuff:
This readme had some outdated information regarding how image queries are handled when the detector's inference pod is not online. I have updated it. I have tested this functionality today, so my description should be accurate.

One thing that is still not clear to me: if I do not configure my detector in `edge-config.yaml`, what is the default behavior? A description of the default behavior might be nice to add to that yaml and/or this readme.


tyler-romero:
The default behavior is the same as the default edge config:
```
edge_inference_configs: # These configs define detector-specific behavior and can be applied to detectors below.
  default: # Return the edge model's prediction if sufficiently confident; otherwise, escalate to the cloud.
    enabled: true
    always_return_edge_prediction: false
    disable_cloud_escalation: false
```
An entry that looks like this will automatically be created for the targeted detector.

tyler-romero (approved):

PR #183, opened 2025-02-17
Combine primary and OODD predictions

f-wright:
This is the final PR to get OODD inference running on the edge! It only adds the combination of the primary and OODD predictions, reducing the confidence of the primary prediction with the OODD inference result. 

Prior PRs pulled OODD models into the edge model database (#175) and started launching inference pods for OODD (#179).


app/core/edge_inference.py
```diff
@@ -51,6 +51,71 @@ async def submit_image_for_inference(inference_client_url: str, image_bytes: byt
...
+    """
+    primary_num_classes = get_num_classes(primary_response)
+
+    primary_output_dict = parse_inference_response(primary_response)
+    logger.debug(f"Primary inference server response: {primary_output_dict}.")
+    oodd_output_dict = parse_inference_response(oodd_response)
+    logger.debug(f"OODD inference server response: {oodd_output_dict}.")
+
+    combined_output_dict = adjust_confidence_with_oodd(primary_output_dict, oodd_output_dict, primary_num_classes)
+    logger.debug(f"Final inference result: {combined_output_dict}.")
```
tyler-romero, line 66:
```suggestion
    logger.debug(f"Combined (primary+oodd) inference result: {combined_output_dict}.")
```

app/core/edge_inference.py
```diff
@@ -51,6 +51,71 @@ async def submit_image_for_inference(inference_client_url: str, image_bytes: byt
...
+        raise ValueError(
+            "Can't get number of classes from inference response with no predictions or multi_predictions."
+        )
+
+
+def adjust_confidence_with_oodd(primary_output_dict: dict, oodd_output_dict: dict, num_classes: int) -> dict:
+    """
+    Adjust the confidence of the primary result based on the OODD result.
+
+    NOTE: This is a duplication of the cloud inference result OODD confidence adjustment logic. Changes should not be
```
tyler-romero, line 95: Can you add a link to the code this duplicates? (And vise versa from there to here?)

app/core/edge_inference.py
```diff
@@ -51,6 +51,71 @@ async def submit_image_for_inference(inference_client_url: str, image_bytes: byt
...
+    if multi_predictions is not None:
+        # multiclass or count case
+        return len(multi_predictions["probabilities"][0])
+    elif predictions is not None:
+        # binary case
+        return 2
+    else:
+        raise ValueError(
+            "Can't get number of classes from inference response with no predictions or multi_predictions."
+        )
```
tyler-romero, line 88:
```suggestion
        raise ValueError(
            "Can't get number of classes from inference response with neither predictions nor multi_predictions."
        )
```

tyler-romero (approved):
LGTM, some small nits. Thanks!

app/core/edge_inference.py
```diff
@@ -51,6 +51,71 @@ async def submit_image_for_inference(inference_client_url: str, image_bytes: byt
...
+    """
+    Adjust the confidence of the primary result based on the OODD result.
+
+    NOTE: This is a duplication of the cloud inference result OODD confidence adjustment logic. Changes should not be
+    made here that bring this out of sync with the cloud OODD confidence adjustment logic.
+    """
+    oodd_confidence = oodd_output_dict["confidence"]
+    oodd_label = oodd_output_dict["label"]
+    primary_confidence = primary_output_dict["confidence"]
+    if oodd_confidence is None or primary_confidence is None:
```
CoreyEWood, line 101: At some point I think we might want to make `parse_inference_response` output a pydantic class rather than an arbitrary dictionary, which would make extracting these values more reliable (and eliminate the need to check for Nones that shouldn't be possible). I don't think it needs to be done in this PR but just mentioning it as a future improvement.

app/core/edge_inference.py
```diff
@@ -51,6 +51,71 @@ async def submit_image_for_inference(inference_client_url: str, image_bytes: byt
             raise RuntimeError("Failed to submit image for inference") from e
 
 
+def get_inference_result(primary_response: dict, oodd_response: dict) -> str:
```
CoreyEWood, line 54: I like how clean this function is! Makes it very easy to see what's going on.

CoreyEWood (approved):
This looks good! There's one place I commented on about potentially combining a couple things. Also overall you separating these changes into multiple PRs made reviewing much easier, thanks for doing that.

app/core/edge_inference.py
f-wright: I added more details about its location, but I'm hesitant to add a link to our private cloud repository in the public edge-endpoint repository. It's also implemented separately for the different modes in the cloud, so can't link to a single spot with the implementation.

app/core/edge_inference.py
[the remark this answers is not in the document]
f-wright: You're totally right that the checks for `multi_predictions` and `predictions` are duplicated. I'd prefer to keep it this way though, since we don't need the `num_classes` for the OODD pipelines. Also, currently the output from `parse_inference_response` is nicely tied to a single inference response, without data about the overall detector.

PR #188, opened 2025-03-07
Tag released containers

tomfaulhaber:
This sets up a cleaner tagging architecture so we'll know what's what for testing and customer deployment.

Here's the message I sent to the #edge channel about this:

I'm about to push a PR that will change the way we handle image tags for the edge-endpoint in ECR to make things a bit more systematized. Feedback very welcome.

**The way it is now:** We tag all merges to main (that pass the tests) with two tags: the git-tag and the `:latest` tag. It's also possible to run the same script locally in to push these two tags or a single-platform version with just the git-tag (the default is the former). `setup-ee.sh` will use the `latest` tag by default, but you can override this by setting the `IMAGE_TAG` environment variable.
This has a few problems:
1. It's possible to accidentally create a `latest` tagged image from code that hasn't been merged.
2. All merges to main are automatically picked up by all the edge-endpoints (this might not be so bad)
3. `latest` is a default for lots of Docker tooling making mistakes more likely
4. We don't support multi-platform test images during development.

**The way it will be after this PR:** We will have 4 supported tags for ECR images:
1. **The git-tag**: All pushes to main will be tagged with the git-tag. This tag is persistent, even after new pushes, so it can be used to fix the version an edge device is using. It will also be possible to use `build-push-edge-endpoint-image.sh` to push either a single- or multi-platform version of the image to ECR.
2. **pre-release**: The latest version of the main branch will be automatically tagged `pre-release`.
3. **release**: GHA will have a human approval gate to move the latest build from `pre-release` to `release` (like what we do on prod deployments in zuuul).
4. **latest**: This is applied the same way as `release`. It will no longer be possible to push this tag from a script (though you could still do it by hand). 

`setup-ee.sh` will continue to default to `latest` for the moment.

**The way it will be when we switch to Helm-based deployment**: There will be three tagging changes:
1. **latest** will be deprecated completely. We will no longer tag images with latest and, once we've updated everyone, we will delete the tag from ECR.
2. **dev** is a local-only tag (never pushed to ECR) to enable local development. These will be single-platform images that will be built locally with Docker and pushed into your local k3s deployment.
3. **The git-tag**: We will no longer support pushing single-platform images to ECR with the git-tag.

We will switch to defaulting to the `release` tag for deployments, but you will be able to override this with a Helm value setting. 

**Further ahead**: Fixed image tags (like `release` or `latest`) present an operational problem at scale. We'll need to work out a plan to use specific image tags and get the system to upgrade as appropriate. I have some ideas here, but haven't done a detailed design yet.

Sorry for the big wall of text. I'm anxious to hear any thoughts on this.


deploy/bin/tag-edge-endpoint-release.sh
```diff
@@ -0,0 +1,54 @@
...
+# Check if an argument is provided
+if [ $# -ne 1 ]; then
+    echo "Usage: $0 <new-tag>"
+    exit 1
+fi
+
+NEW_TAG=$1
+
+# Only the pipeline can create releases
+if [[ "$NEW_TAG" == "pre-release" || "$NEW_TAG" == "release" ]]; then
```
tyler-romero, line 26: Should we add "latest" here as well, for now?

tyler-romero (approved):
LGTM!

.github/workflows/pipeline.yaml
```diff
@@ -424,6 +424,9 @@ jobs:
       - name: Build and Push Multiplatform edge-endpoint Image to ECR
         timeout-minutes: 45
         run: ./deploy/bin/build-push-edge-endpoint-image.sh
+      - name: Tag main images with the "pre-release" tag
+        if: github.ref == 'refs/heads/main'
+        run: ./deploy/bin/tag-edge-endpoint-image.sh pre-release
```
CoreyEWood, line 429: Should this be `tag-edge-endpoint-release.sh`? (also for the two references to this script below)

CoreyEWood (commented):
Has the new tagging setup been tested / is there a way to test it before merging?

.github/workflows/pipeline.yaml
tomfaulhaber: You're right! But I think `tag-edge-endpoint-image.sh` is the right name for what I'm doing here so I'm going to rename the file instead of changing the references.

deploy/bin/tag-edge-endpoint-release.sh
tomfaulhaber: Yes. I'll make that change.

tomfaulhaber:
> Has the new tagging setup with actions been tested / is there a way to test it before merging?

There's no real way to test these gated actions so we'll have to wait and see and then submit any fixes.

I did run the script to prove that it (a) can successfully tag and push an image and (b) won't let me make a release tag from my shell prompt.

PR #189, opened 2025-03-11
Use groundlight version 0.22.3 and set `source=EDGE` / `done_processing=True` for edge IQs

CoreyEWood:
Now IQs created on the edge will have `done_processing=True` if they are fully answered on the edge, and `done_processing=False` if they are answered on the edge but also escalated to the cloud. Also, edge results will have `source=EDGE`.


brandon-groundlight (approved):
Looks good, thanks!

PR #190, opened 2025-03-12
Revert "Use groundlight version 0.22.3 and set `source=EDGE` / `done_processing=True` for edge IQs"

CoreyEWood:
Reverts groundlight/edge-endpoint#189


brandon-groundlight (approved):
Good catch on the backwards compatibility issue

PR #191, opened 2025-03-12
Very minimal EE metrics reported to cloud

robotrapta:
This supports the collecting data for most minimal reporting:

- How many edge devices are running
- Are they active or idle?  (When was the last image-query?)

It also adds a minimal HTML status page if you try to open the EE's port in a browser.  This is just a placeholder, but will become useful soon enough.

This also include https://github.com/groundlight/edge-endpoint/pull/208 to get the e2e tests to pass.


deploy/k3s/edge_deployment/edge_deployment.yaml
robotrapta, line 1: Revert most of this change!  Should only be changing a line or two in here.

test/core/test_metadata_cache.py
```diff
@@ -54,23 +58,123 @@ def test_timestamped_ttl_cache():
...
+    cache.suspend_cached_value("key4")
+
+    cache["key4"] = "value4_updated"
+    cache.suspend_cached_value("key4")
+
+    cache.restore_suspended_value("key4")
+    assert cache.get("key4", None) == "value4_updated"
+
+
+def test_timestamped_cache_error_cases():
```
f-wright, line 157: What happens if you suspend the same value twice, without updating it? I think it errors, should that be in this test?

f-wright (commented):
Looks good, I left a couple of comments. I'm hoping to try it out and look at the web page, but something's wrong with my helm setup so figured I'd send over comments now. Will try to run it again tomorrow

Dockerfile
```diff
@@ -105,7 +105,7 @@ RUN rm /etc/nginx/sites-enabled/default
 RUN ln -sf /dev/stdout /var/log/nginx/access.log && \
     ln -sf /dev/stderr /var/log/nginx/error.log
 
-CMD nginx && poetry run uvicorn --workers 8 --host 0.0.0.0 --port ${APP_PORT} --proxy-headers app.main:app
+CMD ["/bin/bash", "-c", "./app/bin/launch-logic-server.sh"]
```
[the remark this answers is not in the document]
robotrapta: This was a mistake, and was causing everything to fail.  It's fixed now - I renamed the launch script to match the Dockerfile.

app/core/deviceid.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    if not os.path.exists(DEVICE_ID_FILE):
+        return None
+    try:
+        with open(DEVICE_ID_FILE, "r") as f:
+            data = json.load(f)
+
+        if "uuid" in data:
+            return data
+        logger.warning(f"Device ID file {DEVICE_ID_FILE} exists but is missing uuid. Generating new one.")
```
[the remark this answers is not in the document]
robotrapta: Reworded

app/metrics/metricreporting.py
```diff
@@ -0,0 +1,88 @@
...
+        headers,  # header_params
+        payload,  # body
+        async_req=False,  # async_req
+        _return_http_data_only=True,  # _return_http_data_only
+    )
+    logger.debug(f"Report edge metrics: {response}")
+
+
+if __name__ == "__main__":
+    logging.basicConfig(level=logging.DEBUG)
```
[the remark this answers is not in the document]
robotrapta: I don't feel a need to.  It doesn't do anything bad, and facilitates a convenient way to test.  The code only ever runs if you run this file from the python CLI.

test/core/test_metadata_cache.py
robotrapta: I'm confused why this is in my PR - I don't think I wrote this code.  And I don't know what it means to "suspend" something from a cache.

f-wright: Looks like this was just some merge weirdness, seems to be gone now.

app/metrics/iqactivity.py
```diff
@@ -0,0 +1,46 @@
...
+        return Path(self.base_dir, name)
+
+
+@lru_cache(maxsize=1)  # Singleton
+def _tracker() -> FilesystemActivityTrackingHelper:
+    """Get the activity tracker."""
+    return FilesystemActivityTrackingHelper(base_dir="/opt/groundlight/edge-metrics")
+
+
+def record_iq_activity(detector_id: str):
```
f-wright, line 33: Why does this take the detector_id? Seems to record any activity on the edge, regardless of detector

robotrapta: Exactly.  There should be a `# TODO: use the detector_id` right underneath this.

PR #192, opened 2025-03-12
Cache detector metadata indefinitely

CoreyEWood:
Previously, the detector metadata cache would expire after 10 minutes. Additionally, we would attempt to refresh it every 30 seconds, but if the SDK call to refresh it failed the cached value would still be cleared. 

This PR does two things:
1) Makes the detector metadata cache have no time limit so that a cached value will stay indefinitely, unless manually removed (or unless `maxsize=1000` detector IDs get stored in the cache, which seems unlikely)
2) Makes the `refresh_detector_metadata_if_needed` function _not_ clear the cached value if it fails to fetch a new value (e.g., due to no internet connection)

It accomplishes these by removing the TTL aspect of the cache and augmenting it with the ability to "suspend" and "restore" a value from the cache. 

Adds lots of tests to verify both the specific behavior of the cache as well as the overall refreshing behavior.


tyler-romero (approved):
LGTM!

honeytung (approved):
LGTM! Great work!

Issue #193, opened 2025-03-13
GPU capacity never came online after install-nvidia-k3s


robotrapta:
**Describe the bug**

**To Reproduce**
1. On a reasonably "mature" ubuntu22 machine, I ran `./deploy/bin/install-k3s-nvidia.sh'.  It spun for a long time, but the GPU capacity never came online - timed out after 10 minutes.  So the setup failed.
2. I could see that the various pods in the `gpu-operator` space were failing to come online.

**Expected behavior**
I'd _like_ if the install magically worked every time, but I recognize that's a tall order.

**Additional context**
The `apt upgrade -y` step took a very long time, because apparently this machine hadn't been upgraded in a while.  I could see it was churning through many many gigabytes of cuda stuff in the process.  After the setup script timed out, `nvidia-smi` failed saying the drivers were (something) which I recognize as commonly requiring a reboot.

The workaround was to reboot the machine.  The nvidia drivers came back.  I ran the install script again, and it finished quite quickly, but complained (incorrectly this time) that the GPU capacity wasn't online.  So that seems like a bug in the test.

PR #194, opened 2025-03-13
(Tiny) Making install-nvidia respond properly on machines with multiple GPUs.

robotrapta:
The install script was checking if the GPU capacity was exactly 1, so it would incorrctly report failure on machines with multiple GPUs even though it was working correctly.  Partially addresses #193


honeytung (approved):
LGTM! Thanks for catching that!

PR #197, opened 2025-03-13
Use buildx for tagging to support multiplatform

tomfaulhaber:


honeytung (approved):
LGTM, thanks for fixing this!

PR #202, opened 2025-03-25
Updating g4-e2e test to use Helm

robotrapta:
Uses `helm-local` for g4 end-to-end test instead of `setup-ee`.  I chose this instead of pushing everything to ECR in order to keep the tests all running in parallel so they're fast instead of serializing this after ECR-push which is after other tests.

Some other good changes along the way:

- Uses API token to get ECR creds instead of ec2 instance profile
- Uses the merge commit, so we're testing what will happen after the merge


tomfaulhaber (approved):
all lgtm!

PR #203, opened 2025-03-25
(redundant) Minimal device metrics main

robotrapta:
This is the same code as https://github.com/groundlight/edge-endpoint/pull/191 but based on `main` so all the tests run.


robotrapta:
No longer needed because deploy-helm got merged to main.

PR #205, opened 2025-03-26
COM-2300: Support clone-free installs from a self-managed Helm repo

tomfaulhaber:
**Note to testers**: The `deploy/README.md` is written as if the PR was already merged. To test the downloadable k3s installation, replace `main` with `clone-free-install` like this:
```
curl -fsSL https://raw.githubusercontent.com/groundlight/edge-endpoint/refs/heads/clone-free-install/deploy/bin/install-k3s.sh | bash -s cpu
```
(but set the `cpu`/`gpu` option correctly for your system).

---
This PR includes:
- GHA to deploy packaged Helm charts to a repo on the `gh-pages` branch
- Consolidated the k3s install into a single shell script that can be run with `curl | bash` that includes both cpu and gpu setup. Cleaned up the old sub-scripts that we used to use there.
- Updated docs to tell users how to install from the internet without cloning

Also, I've been updating the [gh-pages branch](https://github.com/groundlight/edge-endpoint/tree/gh-pages) to act as a Helm repo with a simple landing page and README. You can see the customer view at https://code.groundlight.ai/edge-endpoint/.

### Checklist:
- [x] Create a helm repo using github pages ✅ 2025-03-25
- [x] Fix instructions on README and index.html in gh-pages branch ✅ 2025-04-09
- [x] Figure out chart versioning ✅ 2025-03-26
- [x] Add a GHA to publish charts to the repo ✅ 2025-03-26
- [x] Add the repo during the install-k3s sequence ✅ 2025-03-26
- [x] Create a single-script install-k3s routine that supports `curl -o - | bash -` style installs. ✅ 2025-04-07
- [x] Clean up old files that made up parts of the k3s install ✅ 2025-04-09
- [x] Fix GHA tests to specify gpu/cpu on the k3s install and never to use install-k3s-nvidia
- [x] Add sample file(s): ✅ 2025-04-10
	- [x] `edge-config.yaml` ✅ 2025-04-09
	- [ ] Others?
- [x] Document the install process in the `deploy/README.md` ✅ 2025-04-10


.github/workflows/pipeline.yaml
```diff
@@ -548,8 +548,8 @@ jobs:
       - name: Update GLHub
         run: |
           cd glhub
-          git config --global user.email "roxanne+gh_bot@groundlight.ai"
-          git config --global user.name "edge-glhub-bot"
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
darylhansen, line 552: 👍 . Probably want to update lines 559:560 as well (or remove them? seems duplicated)

darylhansen (commented):
Neat! I'll have more detailed feedback once I've had a chance to try installing it.

.github/workflows/pipeline.yaml
tomfaulhaber: Yes! Thanks. Done.

brandon-groundlight (approved):
This worked pretty well for me. I already had k3s installed on my machine, so just went through the remaining steps then submitted some image queries to it. Only have a couple comments on the docs

On the GH-Pages branch, it says 
> This repository only contains the Helm chart
But should probably be 
> This branch only contains the Helm chart
Similarly where it reads
> The repo is automatically updated

CoreyEWood (approved):
Worked well for me! (other than the k3s installation issues on beebop)

test/integration/test-with-k3s-setup-ee.sh
```diff
@@ -38,6 +38,11 @@ export INFERENCE_FLAVOR="CPU"
 export LIVE_TEST_ENDPOINT="http://localhost:$EDGE_ENDPOINT_PORT"
 export REFRESH_RATE=60 # not actually different than the default, but we may want to tweak this
 
+# Compute the image tag name before we muck with the config file so we get
+# the tag that will correspond to the current commit so it can match the image
+# that was built and pushed to ECR
+export IMAGE_TAG=$(./deploy/bin/git-tag-name.sh)
```
tyler-romero, line 44: Does this mean that if we run this test twice, the second time it wont grab the same image? I suppose this doesnt matter if we're using this only in CICD.

tyler-romero (approved):
LGTM!

test/integration/test-with-k3s-setup-ee.sh
tomfaulhaber: It should grab the same image both times. The issue is that the way Roxanne set up the test, it modifies the edge-config file in place for the duration of the test. If we run the git-tag-name.sh after that we get a tag name that has never existed before and, therefore, there can't really be an image for this. So I want to be sure that I grab the tag first.

She restores the original file after the test, so a second run should get the same tag.

This isn't the most robust thing. But we're going to deprecate this whole `setup-ee.sh` mechanism RSN, so I'm not too worried about it. For the helm version of the test, I use a file in `/tmp`  to avoid this problem and just because it seems better.

tomfaulhaber:
@brandon-groundlight 

>On the GH-Pages branch, it says

    This repository only contains the Helm chart
    But should probably be
    This branch only contains the Helm chart
    Similarly where it reads
    The repo is automatically updated

When I use the word repo there, I mean the Helm repository (which is stashed in the gh-pages branch of the github repository). So the word "branch" would not be correct there, but I agree that it's confusing and will clarify. Thanks for calling it out.

deploy/README.md
```diff
@@ -8,70 +8,62 @@ The instructions below are fairly opinionated, optimized for single-node cluster
...
-2. [Set up a local Kubernetes cluster with k3s](#setting-up-single-node-kubernetes-with-k3s).
-3. [Set your Groundlight API token](#set-the-groundlight-api-token).
-4. [Set up your shell to run Helm](#setting-up-for-helm).
-5. [Install the Edge Endpoint with Helm](#installation-with-helm).
-6. [Confirm that the Edge Endpoint is running](#verifying-the-helm-installation).
+1. [Set up a local Kubernetes cluster with k3s](#setting-up-single-node-kubernetes-with-k3s).
+2. [Set your Groundlight API token](#set-the-groundlight-api-token).
+3. [Set up to use the Helm package manager](#setting-up-for-helm).
+4. [Install the Edge Endpoint with Helm](#installation-with-helm).
+5. [Confirm that the Edge Endpoint is running](#verifying-the-helm-installation).
```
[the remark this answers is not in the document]
tomfaulhaber: Fixed. Thanks!

PR #207, opened 2025-03-26
Adding diagnostics to failed CI/CD runs

robotrapta:
Adding diagnose-failure steps to the validate-* workflow jobs, which show what happened inside kubernetes on a failure.

### DIFF
--- a/.github/workflows/pipeline.yaml
+++ b/.github/workflows/pipeline.yaml
@@ -148,19 +148,8 @@ jobs:
 
       - name: Diagnose failure
         if: failure()
-        env:
-          # Namespace is set in validate_setup_ee.sh
-          NAMESPACE: validate-setup-ee
         run: |
-          set -x
-          set +e
-          df -h
-          kubectl config set-context --current --namespace=$NAMESPACE
-          kubectl get all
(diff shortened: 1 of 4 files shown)


.github/workflows/pipeline.yaml
```diff
@@ -107,8 +107,15 @@ jobs:
         run: docker stop ${{ steps.start_container.outputs.container_id }}
 
   validate-setup-ee:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-22.04-8core
+    # We'd prefer like to use ubuntu-latest-4-cores, but they're not working.
+    # Standard 2 core doesn't have enough disk space.  (73GB total, 21GB available)
+    # 8core has tons (290GB, 243GB available)
     steps:
+      - name: Show free disk space
```
tomfaulhaber, line 115: You've got this in the error script. Do we really want it on every run?

tomfaulhaber (approved):

.github/workflows/pipeline.yaml
robotrapta: Yeah we don't need it here.  Good catch.

PR #208, opened 2025-03-27
Integration tests uses fresh code instead of `:latest`

robotrapta:
Previously many of the full (k3s) CI/CD tests used `:latest` images from ECR.  This PR changes them to all use the current build.  This requires changing the order so that we always build & push to ECR at the beginning of any pipeline run, and then use that image in all subsequent tests.


deploy/bin/build-push-edge-endpoint-image.sh
```diff
@@ -30,7 +31,8 @@ aws ecr get-login-password --region ${ECR_REGION} | docker login \
                   --username AWS \
                   --password-stdin  ${ECR_URL}
 
-# TODO: delete this!!!
+# TODO: delete this!!!  (Why?  Because we want to standardize on the multiplatform build I think.
```
robotrapta, line 34: I dunno - what do you think?

test/integration/test-with-k3s-helm.sh
```diff
@@ -51,20 +51,20 @@ fi
...
+echo "INFERENCE_FLAVOR: $INFERENCE_FLAVOR"
+echo "DEPLOYMENT_NAMESPACE: $DEPLOYMENT_NAMESPACE"
+echo "IMAGE_TAG: $IMAGE_TAG"
 helm install -n default ${HELM_RELEASE_NAME} deploy/helm/groundlight-edge-endpoint \
     --set groundlightApiToken=$GROUNDLIGHT_API_TOKEN \
     --set inferenceFlavor=$INFERENCE_FLAVOR \
     --set edgeEndpointPort=$EDGE_ENDPOINT_PORT \
     --set namespace=$DEPLOYMENT_NAMESPACE \
     --set edgeEndpointTag=$IMAGE_TAG \
-    --set imagePullPolicy=Never \
```
f-wright, line 67: I don't totally know what the different image pull policies do -- what's the effect of removing this?

f-wright (approved):
LGTM!

test/integration/test-with-k3s-helm.sh
robotrapta: The image-pull policy tells k8 when it should go to the image repo (ECR for us) and try to download the image it's being told to use.  `never` is a strange case - it has little use in the real world, because it generally doesn't work.  It only works if you somehow inject the image into k8 yourself before creating the deployment.  Which is what we do for `:dev` builds where we short-circuit ECR entirely and just push it straight into k8 and not to ECR.  That's what was going on before.  With that specified, the test failed, because it needs to ge tthe image from ECR.  Without it, things just work normally.

.github/workflows/pipeline.yaml
```diff
@@ -490,21 +486,38 @@ jobs:
...
       - test-with-k3s-helm
+      - validate-setup-ee
+      - validate-setup-helm
+      - G4-end-to-end
+    steps:
+      - name: declare success
+        run: echo "Main suite of tests all pass."
+
+  tag-pre-release:
+    runs-on: ubuntu-22.04
```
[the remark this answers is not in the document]
robotrapta: No reason.

PR #210, opened 2025-03-28
Automatically set the pull policy to "Never" for images tagged "dev"

tomfaulhaber:


deploy/helm/groundlight-edge-endpoint/templates/_helpers.tpl
```diff
@@ -88,6 +88,29 @@ Create the name of the service account to use
...
+  Determine the correct pull policy to use for each container type. If it is 
+  a dev tag, we use "Never" to avoid pulling from the registry. Otherwise,
+  we use the global pull policy.
+*/}}
+{{- define "groundlight-edge-endpoint.edgeEndpointPullPolicy" -}}
+{{- $tag := include "groundlight-edge-endpoint.edgeEndpointTag" . -}}
+{{- if eq $tag "dev" -}}
+Never
+{{- else -}}
+{{- default "IfNotPresent" .Values.imagePullPolicy -}}
```
f-wright, line 101: Is this the default we want? Seems like the default is "Always" in `values.yaml` right now, should the default to be the same here? Same comment below

f-wright (approved):
LGTM, thanks!

deploy/helm/groundlight-edge-endpoint/templates/_helpers.tpl
tomfaulhaber: "IfNotPresent" is the K8s default and I thought that it made sense to use that if the pull policy is set to ""

PR #211, opened 2025-03-31
Additional edge metrics

f-wright:
Leo added some basic edge metrics in https://github.com/groundlight/edge-endpoint/pull/191. This PR adds more metrics! It has a million commits since it was initially based off of the minimal-device-metrics branch, which was in turn based off the helm-deploy branch.

Some photos of the new status page with all the new metrics are below. I also verified that the metrics make it to opensearch properly ([log](https://search-positronix-test-drs7cdy6j5ftkdjzjfyncavkdq.us-west-2.es.amazonaws.com/_dashboards/app/discover#/doc/192040c4-01e8-4beb-94cb-a45f62ef0166/logstash-prod-2025.04.03?id=MmVw_ZUBfTiQ178Z8DW1)), but would be interested in feedback on whether we want detector/pod specific metrics parsed into different fields like this -- could instead send them as lists so opensearch puts them in a single field if that'd be better.

image
image


app/metrics/iq_activity.py
```diff
@@ -0,0 +1,208 @@
+"""Uses the filesystem to track various metrics about image-query activity. Tracks iqs, escalations,
+and audits for each detector, as well as iqs submitted to the edge-endpoint as a whole.
+
+Filesystem structure:
+/opt/groundlight/edge-metrics/
+    detectors/
+        <detector_id1>/
+            iqs
+            iqs_YYYY-MM-DD_HH   <--- Current hour
```
f-wright, line 9: I'd rather not break up our activity tracking by hour like this, but doing a simple counter seemed like the most efficient way to use the filesystem based activity tracking to get an idea of rates of IQ submission and escalations. Highly open to other suggestions though

app/metrics/iqactivity.py
f-wright, line 1: Renamed to match the pattern in the rest of the repo

app/status_monitor/static/status.html
f-wright, line 1: These changes are all cursor, but I tested it and seems to work fine

app/metrics/iq_activity.py
```diff
@@ -0,0 +1,208 @@
...
+
+def record_activity(detector_id: str, activity_type: str):
+    """Records an activity from a detector. Currently supported activity types are:
+    - iqs
+    - escalations
+    - audits
+    """
+    supported_activity_types = ["iqs", "escalations", "audits"]
+    if activity_type not in supported_activity_types:
+        logger.error(
```
robotrapta, line 108: I think I'd raise an exception here.  The only way this happens is through a coding bug / typo.  Logging an error makes it likely somebody doesn't notice, and the system continues along half-working.  For production systems, it's better to fail loudly and for the engineer to fix it instead of keep a half-working system online.  Especially if we know users can't trigger the error.

app/metrics/iq_activity.py
```diff
@@ -0,0 +1,208 @@
...
+    if activity_type == "iqs":
+        # Record the time of the last IQ, included since this file was used in the first version of EE metrics
+        f = _tracker().file("last_iq")
+        f.touch()
+
+        _tracker().increment_counter_file(activity_type)
+        _tracker().increment_counter_file(f"{activity_type}_{current_hour}")
+
+
+def last_activity_time() -> str:
```
robotrapta, line 129: Organizationally, I feel like these methods which return metrics should be grouped somehow.  Maybe they're all class methods in a class `RetrieveMetrics` or something.  Just to show that they belong together or something, and are different from something like `clear_old_activity_files`

app/metrics/iq_activity.py
```diff
@@ -0,0 +1,208 @@
...
+from functools import lru_cache
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+
+class FilesystemActivityTrackingHelper:
+    """Helper class to support tracking image-query activity using the filesystem.
+    This is just a skeleton and only supports timestamps right now.  But
+    we will expand this to support counting metrics, etc."""
```
robotrapta, line 40:
```suggestion
    """
```
You did that, rigth?

app/metrics/iq_activity.py
```diff
@@ -0,0 +1,208 @@
...
+        else:
+            f = self.file(name)
+
+        if not f.exists():
+            f.touch()
+            f.write_text("1")
+            return
+
+        read_total = int(f.read_text())
+        f.write_text(str(read_total + 1))
```
robotrapta, line 84: There's clearly a race-condition here where two processes both try to update the file at once, and the sum goes up by 1 instead of 2.  Maybe we don't care.  It will only ever under-count.  We should certainly make a note of this possibility.

robotrapta: If we wanted to fix this, there are a few ways we could do it:
- Use a database like sqlite.  Seems heavy.
- Use lockfiles to allow one process to claim ownership and make others wait.  A lot of work to do right.  DB is probably easier.
- Switch to append-only semantics.  Appending is atomic across processes.  We could say each time we increment that we just add a `.` character to the file, and then the count is the filesize.  This is probably pretty easy.  The key downside I can think of is that these files could in theory grow large.  But we clean them up regularly?  It's probably less total I/O activity than the current approach.

app/metrics/iq_activity.py
```diff
@@ -0,0 +1,208 @@
...
+    current_hour = datetime.now().strftime("%Y-%m-%d_%H")
+    last_hour = (datetime.now() - timedelta(hours=1)).strftime("%Y-%m-%d_%H")
+
+    detector_metrics = {}
+
+    for activity_type in ["iqs", "escalations", "audits"]:
+        f = _tracker().detector_file(detector_id, activity_type)
+        last_activity = _tracker().get_last_file_activity(f)
+        last_activity = last_activity.isoformat() if last_activity else "none"
+        total_activity = int(f.read_text()) if f.exists() else 0
```
robotrapta, line 170: The `int(f.read_text()) if f.exists() else 0` should definitely be in a method parallel to `increment_counter_file` to allow the kind of refactoring I suggested that solves the race condition.  And to avoid the abstraction used inside `increment` from leaking out in lots of places.

app/main.py
```diff
@@ -53,7 +55,8 @@ async def startup_event():
 
     logging.info(f"edge_config={app.state.app_state.edge_config}")
 
-    scheduler.add_job(report_metrics_to_cloud, "interval", seconds=3600)
+    scheduler.add_job(report_metrics_to_cloud, "interval", seconds=ONE_HOUR_IN_SECONDS)
+    scheduler.add_job(clear_old_activity_files, "interval", seconds=ONE_HOUR_IN_SECONDS)
```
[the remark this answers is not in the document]
robotrapta: I see - we don't care.

app/metrics/system_metrics.py
```diff
@@ -0,0 +1,67 @@
...
+import psutil
+from kubernetes import client, config
+
+logger = logging.getLogger(__name__)
+
+
+def get_cpu_usage() -> str:
+    """Returns the percentage of total CPU used."""
+    percent = psutil.cpu_percent(interval=1)
+    return f"{percent}%"
```
robotrapta, line 13: To include units or not?  There's a tradeoff here between clarity and downstream analysis.  Having the units (`%` or `GB`) in the value makes them ambiguous.  But foils processes like OpenSearch which will want to interpret these as numbers.  Since the goal of this stuff ultimately is machine-readability, I think we should have the return values be floats, and put the units in the dict keys.  e.g. `"cpu_usage_pct": 57.0`.  Doing that either means you need to carefully coordinate the units in the method names & key names, e.g. `get_cpu_usage_pct` or if you're feeling ambitious you could return a tuple like `(57.0, "pct")` and update SafeDict to know how to deal with it.

app/metrics/system_metrics.py
```diff
@@ -0,0 +1,67 @@
...
+def get_percentage_memory_used() -> str:
+    """Returns the percentage of total memory used."""
+    percent = psutil.virtual_memory().percent
+    return f"{percent}%"
+
+
+def get_memory_available() -> str:
+    """Returns the amount of memory available in GB."""
+    total = psutil.virtual_memory().total
+    return f"{total / (1024 ** 3):.2f} GB"
```
robotrapta, line 25: I think raw bytes is probably what we want here.  The skeleton status page is a little deceptive in that it makes us think these are for human consumption, but really we want these numbers going into a tracking database, and have a better UI summarizing them.

robotrapta (commented):
This is great, and I'm tempted to give it a thumbs up knowing you'll respond to the the suggestions.  But I have a number of changes I'd like to see to make it better that I think collectively matter enough for me to send it back for a revision.

app/status_monitor/static/status.html
robotrapta: vibe-coding FTW.

app/main.py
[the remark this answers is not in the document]
f-wright: Yeah, I'm keeping more history than we need.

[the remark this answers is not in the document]
f-wright: Yep, I keep more history than we need.

app/metrics/metric_reporting.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def metrics_payload(self) -> dict:
+        """Returns a dictionary of metrics to be sent to the cloud API."""
+        device_info = SafeMetricsDict()
+        device_info.add("device_id", lambda: deviceid.get_deviceid_str())
+        device_info.add("device_metadata", lambda: deviceid.get_deviceid_metadata_dict())
+        device_info.add("now", lambda: datetime.now().isoformat())
+        device_info.add("cpucores", lambda: os.cpu_count())
+        device_info.add("inference_flavor", lambda: system_metrics.get_inference_flavor())
+        device_info.add("cpu_usage_pct", lambda: system_metrics.get_cpu_usage_pct())
```
tyler-romero, line 64: nit: a standard way to say usage_pct is `utilization`

tomfaulhaber (approved):
LGTM!

PR #212, opened 2025-03-31
Improve exception string creation for model info fetching errors

CoreyEWood:
Previously, the `response.json()` calls would produce exceptions when the response could not be turned into JSON (e.g., in the case of a 401 error due to an invalid API token). Then the error while updating the model would be reported as `requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`, obscuring the real issue.

This PR adds handling for this case and tries to provide detail about the response in any scenario.

Now the 401 error will be correctly reported:
`fastapi.exceptions.HTTPException: 401: Failed to fetch model info for detector 'det_2mIZTxUnJDhhUtJbUXavUevW15K'. Received error: Could not authenticate with any method.`


f-wright (approved):
LGTM!

PR #216, opened 2025-04-02
Make use of the `from_edge` and `done_processing` fields (take two)

CoreyEWood:
Trying again to implement what [this PR did](https://github.com/groundlight/edge-endpoint/pull/189), except without requiring the client submitting to the edge endpoint to have an up-to-date SDK version.

Any Result produced on the edge will have `from_edge=True`. Additionally, results produced in the cloud will never have `from_edge=True`, so this field can decisively be used to determine if a result was produced on the edge.

An IQ will have `done_processing=True` when polling for an answer at a later point (based on the query's ID) has no chance of producing a better answer. This is the case when a query was answered on the edge and was not escalated to the cloud.

To ensure that setting these fields does not break things for users still on old SDK versions, I wrote a script to iterate through previous versions of the Groundlight Python SDK and submit an image query to an edge endpoint running this branch's code. I tested the previous 30 versions and all succeeded, going back to `0.15.0` (released on April 1, 2024). I included this rough testing script in case we want to use it in the future.


app/api/routes/image_queries.py
```diff
@@ -197,6 +198,8 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
                         metadata=generate_metadata_dict(results=results, is_edge_audit=True),
                         image_query_id=image_query.id,  # We give the cloud IQ the same ID as the returned edge IQ
                     )
+                    # Not done processing because the associated IQ in the cloud could get a better answer
+                    image_query.done_processing = False
```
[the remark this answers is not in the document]
CoreyEWood: For now at least, `done_processing` is edge specific and doesn't relate to patience time. It's true if, for whatever reason, the image query is done escalating / will no receive no new results - right now, this would be if it either got a confident answer on the edge or if cloud escalation is disabled.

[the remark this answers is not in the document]
timmarkhuff: Since the audit is invisible to the user, I think we should consider it to be `done_processing=True` here. 

Also, I believe the audit could go beyond the patience time, right? If so, that's another reason to disregard it here.

app/core/utils.py
```diff
@@ -54,6 +55,7 @@ def create_iq(  # noqa: PLR0913
     :param result_value: The predicted value.
     :param confidence: The confidence of the predicted value.
     :param confidence_threshold: The confidence threshold for the query.
+    :param is_done_processing: Whether this is the final answer for the query.
```
timmarkhuff, line 58: Should we go a bit further here and say "the best answer within patience time"? Brandon  can correct me if this isn't the right terminology, but I think it is.

app/api/routes/image_queries.py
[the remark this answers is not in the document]
CoreyEWood: That makes sense, I'm down to make `done_processing=True` for the audit case. 

And true, the audit could go beyond the patience time - though I think that's true for any cloud result.

app/core/utils.py
```diff
@@ -102,6 +105,7 @@ def _mode_to_result_and_type(
         result = BinaryClassificationResult(
             confidence=confidence,
             source=source,
+            from_edge=True,
```
timmarkhuff, line 108: Are all IQs coming from an edge endpoint set with `from_edge=True`?

I thought that if the edge endpoint escalates to cloud, we would see `from_edge=False`, but I am not seeing how that works. Maybe that goes through a different code path I am not seeing?

CoreyEWood: Only results that are actually produced on the edge will have `from_edge=True`. 

When the edge endpoint escalates to the cloud, it returns the result of the escalation, which will be an IQ created in the cloud. All results in the cloud will have `from_edge=False`, so when the edge returns a query from the cloud, its result will have `from_edge=False`.

app/core/utils.py
brandon-wada: Wait defines how long the client will poll for an answer. Patience defines how long Groundlight will actively continue to escalate the query.
In a really annoying turn, one could get a better answer within patience time if a customer added a GT label within patience time, but that may not be the answer returned here. You could use "Whether Groundlight has completed all planned escalations" to be technically correct

honeytung (approved):
LGTM! Thanks for making this change! It is great that we can now check the `from_edge` flag to see if the IQ exist on cloud instead of trying to do a request to the cloud.

Do you also run the test to see if old client SDK will be affected by this change?

PR #217, opened 2025-04-03
Fix image tagging to work on the multiplatform manifest

tomfaulhaber:


honeytung (approved):
LGTM! Thanks for fixing that bug!

PR #219, opened 2025-04-03
Never write single platform images to ECR

tomfaulhaber:


honeytung (approved):
Thanks! Looks good and the tests are not failing as well!

PR #222, opened 2025-04-04
Make test workflow reusable to trigger edge tests for new inference server images

f-wright:
This PR splits out building the EE image and the helm k3s tests into reusable workflows, which can be called outside of this repo. The goal for this update is to be able to trigger the EE tests when we build a new inference image. Since this is a public repo, this does mean that the workflow could be called by any other repo, but I'm not really concerned about that since it requires the secrets to be passed in. 

`test-with-k3s-helm` validates that the inference deployments come online and runs full integration tests using them, so I think it's the right selection to test a new inference image. `test-with-k3s-setup-ee` would do the same, but it wasn't immediately clear how to specify the inference image, and since we're switching over to helm I think it's fine to just run that one. 

The tests on this PR pass and were run on the correct images. This [zuuul PR](https://github.com/positronix-ai/zuuul/pull/4883) will trigger these new workflows when we build a new inference image, and has been tested on this branch. Some modifications will need to be made to switch it to the main branch, and it will need to be merged after this one goes in.


.github/workflows/pipeline.yaml
```diff
@@ -487,10 +506,20 @@ jobs:
...
         # This is actually only taking about 4 minutes right now.
         timeout-minutes: 45
-        run: ./deploy/bin/build-push-edge-endpoint-image.sh
+        run: |
+          echo "PWD: $(pwd)"
+          echo "LS: $(ls -a)"
+          echo "cd .."
+          cd ..
+          echo "PWD: $(pwd)"
+          echo "LS: $(ls -a)"
```
robotrapta, line 521: I'm assuming this is just debugging stuff while you figure it out?

.github/workflows/pipeline.yaml
```diff
@@ -487,10 +506,20 @@ jobs:
           mask-password: "true"
       - name: Check out code
         uses: actions/checkout@v4
+        with:
+          # Check out the edge-endpoint repo, not the repo that triggered the workflow
```
robotrapta, line 510:
```suggestion
          # Check out the edge-endpoint repo, not the repo that triggered the workflow
          # Checks out into the subdir `edge-endpoint` so we need to adapt our paths accordingly. 
```
Yes?

robotrapta (commented):
Looks reasonable to me, but I've never done anything like this.  Have @darylhansen take a look?

.github/workflows/pipeline.yaml
f-wright: Oh yeah, sorry I initially sent this PR out prematurely. I thought I could use repository dispatch, but then realized that zuuul won't wait for the result of the workflow. This PR is back in draft mode and I'm working on a new way.

.github/workflows/pipeline.yaml
```diff
@@ -7,6 +7,25 @@ on:
     branches:
       - main
     types: [opened, synchronize, reopened]
+  workflow_call:
```
darylhansen, line 10: I think this is not what you want. It's going to get really confusing having the same thing be triggered by itself, but also be callable.

What I recommend is two files:
- `pipeline.yaml`, which has `on: workflow_call` at the top, and just does the reusable parts
- `cicd.yaml`, which has `on: push, pull_request, workflow_dispatch`, runs the jobs you labeled with `if: github.event_name != 'workflow_call'`, and has a job that runs the callable workflow.

Happy to talk it through in more detail!

darylhansen (commented):
> I verified that these changes didn't make any difference to how the GHA pipeline ran on this PR. This https://github.com/positronix-ai/zuuul/pull/4883 will trigger this workflow when we build a new inference image, but this PR needs to go in first.

For what it's worth, if you want to test this before submitting you can change this line in the other PR
```
uses: groundlight/edge-endpoint/.github/workflows/pipeline.yaml@main
```
to end with `@gha-for-zuuul`, and it'll use this branch. Just make sure to change it back once it's merged.

.github/workflows/pipeline.yaml
f-wright: I considered this, but thought it might get more complicated since basically all of the jobs rely on `build-push-edge-endpoint-multiplatform`, which also needs to be run on workflow_call. I wouldn't want to make the CICD yaml wait on the whole pipeline.yaml job, since `test-with-k3s-helm` can take a while and can be run in parallel with the other long running jobs. 

Maybe the solution is 3 files, moving `build-push-edge-endpoint-multiplatform` into it's own? Or is there a better way to handle that dependency that you know of?

darylhansen (approved):
Nice and clean!

PR #224, opened 2025-04-09
Handle the fact that ELBs change IP addresses

tomfaulhaber:
This should deal with the fallback upstream URL in our edge-endpoint breaking after a while because the ELB changes its IP address. 

I only made this change for the helm version since making it for the `setup-ee.sh` version would have required us to do a bunch of duplication and/or scripting (so that the docker tests would continue to pass). Since we're deprecating that version, I figured I wouldn't create that mess.

I think the python SDK may have a similar problem for long-running programs (like the edge-endpoint app), but I need to explore that further.

### DIFF
--- a/deploy/helm/groundlight-edge-endpoint/files/nginx.conf
+++ b/deploy/helm/groundlight-edge-endpoint/files/nginx.conf
@@ -2,6 +2,8 @@ events {
     worker_connections 256;
 }
 
+
+
 http {
     # Add "NGINX: " prefix to the log entries to distinguish them from other logs.
     # Follows NGINX's default "combined" log format.
@@ -36,9 +38,15 @@ http {
         }
 
         location @fallback {
-            # Fallback to the cloud API server
+            # This is the default resolver for Kubernetes. We need to set it explicitly
+            # and put the fallback server in a varible to get nginx to resolve it regularly.
+            # This is important when going to an AWS ELB because the IPs change.
+            resolver 10.43.0.10 valid=30s;
             {{- $parsedURL := .Values.upstreamEndpoint | urlParse }}
-            proxy_pass {{  index $parsedURL "scheme" }}://{{ index $parsedURL "host" }};
+            set $upstreamEndpoint {{  index $parsedURL "scheme" }}://{{ index $parsedURL "host" }};
+
+            # Fallback to the cloud API server
+            proxy_pass $upstreamEndpoint;
         }
 
     }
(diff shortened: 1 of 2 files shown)


deploy/helm/groundlight-edge-endpoint/files/nginx.conf
```diff
@@ -2,6 +2,8 @@ events {
     worker_connections 256;
 }
 
+
+
```
darylhansen, line 6: nit: extra newlines

deploy/helm/groundlight-edge-endpoint/files/nginx.conf
```diff
@@ -36,9 +38,15 @@ http {
         }
 
         location @fallback {
-            # Fallback to the cloud API server
+            # This is the default resolver for Kubernetes. We need to set it explicitly
+            # and put the fallback server in a varible to get nginx to resolve it regularly.
+            # This is important when going to an AWS ELB because the IPs change.
```
darylhansen, line 43: 🫠  makes sense

darylhansen (approved):

deploy/helm/groundlight-edge-endpoint/files/nginx.conf
```diff
@@ -36,9 +38,15 @@ http {
         }
 
         location @fallback {
-            # Fallback to the cloud API server
+            # This is the default resolver for Kubernetes. We need to set it explicitly
+            # and put the fallback server in a varible to get nginx to resolve it regularly.
```
brandon-groundlight, line 42: I would never have realized this is how to get around the caching in proxy_pass. Nice

brandon-groundlight (approved):
I feel like I would have struggled pretty hard to figure this out myself. This is great

deploy/helm/groundlight-edge-endpoint/files/nginx.conf
tomfaulhaber: Cleaned up. Thx!

deploy/helm/groundlight-edge-endpoint/files/nginx.conf
```diff
@@ -36,9 +36,15 @@ http {
         }
 
         location @fallback {
-            # Fallback to the cloud API server
+            # This is the default resolver for Kubernetes. We need to set it explicitly
+            # and put the fallback server in a variable to get nginx to resolve it regularly.
+            # This is important when going to an AWS ELB because the IPs change.
+            resolver 10.43.0.10 valid=30s;
```
robotrapta, line 42: This makes me pretty nervous, TBH.  I don't think we have any guarantees that `10.43.0.10` is where coredns will land, do we?  Can we explicitly configure it when setting up k3s?  Or create a k8 service which points to it or something?

PR #225, opened 2025-04-09
Don't include `results` in metadata if it would pass the max allowed size

CoreyEWood:
Previously, with ~7+ ROIs in the `results` dict, cloud escalations would fail because the metadata exceeded the maximum allowed size. This ensures we don't exceed the size and restricts the contents of the `results` data if necessary.

The max size is defined [here](https://github.com/groundlight/python-sdk/blob/67c8d291da33f7b42302fa49c08bc73745210f68/src/groundlight/client.py#L342) in the SDK. Unfortunately the number is hard-coded in, so if it changes in the SDK we'll have to manually update it here.

### DIFF
--- a/app/core/utils.py
+++ b/app/core/utils.py
@@ -141,6 +141,17 @@ def safe_call_sdk(api_method: Callable, **kwargs):
         raise ex
 
 
+def _size_of_dict_with_field(initial_dict: dict[str, Any], new_data_key: str, new_data_value: Any) -> int:
+    """
+    Returns the size, in # of bytes (assuming all ASCII characters), of the provided dict if it includes the provided
+    key/value pair.
+    """
+    combined_dict = initial_dict.copy()
+    combined_dict[new_data_key] = new_data_value
+    combined_json = json.dumps(combined_dict)
+    return len(combined_json)
+
+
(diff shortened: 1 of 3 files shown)


app/core/utils.py
```diff
@@ -138,6 +141,30 @@ def safe_call_sdk(api_method: Callable, **kwargs):
...
+    metadata_dict = {}
+
+    if is_edge_audit:
+        metadata_dict["is_edge_audit"] = True  # This metadata will trigger an audit in the cloud
+
+    metadata_dict["edge_result"] = results
+    metadata_json = json.dumps(metadata_dict)
+    size_bytes = len(metadata_json)
+    if size_bytes > METADATA_SIZE_LIMIT_BYTES:
+        metadata_dict.pop("edge_result")
```
paulina-positronix, line 159: Would really love to preserve the final answer and confidence, if it's not too difficult. It's the ROIs that are the problem if there are too many of them, can we replace any rois subfield with a string that says something like f"{len(rois)} rois were detected"?

paulina-positronix: Although in bounding box mode the answer itself is ROIs?

honeytung (approved):
Looks good! Thanks for the comprehensive tests!

PR #230, opened 2025-04-21
Configurable log level and move noisy logs to DEBUG

tomfaulhaber:
This PR does two things:
1. Make it so you can configure the log level for the apps in the helm deploy step with `--set logLevel=DEBUG` or whatever.
2. Move some noisy logs from `INFO` to `DEBUG` so the logs will focus more on what's really happening.

When the system is doing nothing, the only noisy logs come from health checks.


tyler-romero (approved):
LGTM, thanks!

PR #231, opened 2025-04-22
Add Jetson Support in Installation Script

honeytung:
This PR adds a new option for Jetson install. It is similar to `gpu` installation but ignores some GPU checks that are not supported in Jetson platforms.

Note: Jetson installation is currently experimental and we will update the README to provide detailed installation guide when it is ready.


tomfaulhaber (approved):
LGTM

PR #234, opened 2025-04-29
Adds Support for Multiclass Mode

honeytung:
WIP


CoreyEWood (commented):
Still taking a look at everything, but could you also add a test for parsing a multiclass response in `test_parse_inference_response.py`? That'll help make sure we handle the raw inference response of a multiclass pipeline properly.

app/core/utils.py
```diff
@@ -119,8 +121,15 @@ def _mode_to_result_and_type(
...
-        raise NotImplementedError("Multiclass functionality is not yet implemented for the edge endpoint.")
-        # TODO add support for multiclass functionality.
+        if mode_configuration is None:
+            raise ValueError("mode_configuration for MultiClass detector shouldn't be None.")
+        multi_class_mode_configuration = MultiClassModeConfiguration(**mode_configuration)
+        result_type = ResultTypeEnum.multi_classification
+        result = MultiClassificationResult(
+            confidence=confidence,
+            source=source,
+            label=multi_class_mode_configuration.class_names[result_value],
```
CoreyEWood, line 131: For some reason indexing into the list to get the label feels so sketchy to me but I know this is how we do it so I don't have an issue here, just wanted to express that.

pyproject.toml
```diff
@@ -10,7 +10,7 @@ APScheduler = "3.10.4"
 cachetools = "^5.3.1"
 fastapi = "^0.115.0"
 framegrab = "^0.5.0"
-groundlight = ">=0.21.0, <0.22.0"
+groundlight = ">=0.22.0, <0.22.4"
```
CoreyEWood, line 13: Version changes on the edge make me nervous ever since I temporarily broke things for older SDK version users a couple months ago. It seems like this should be okay, and I tested getting multiclass responses back using versions 0.19.0 and 0.21.0. Definitely worth a double check all around though.

CoreyEWood (approved):
The multiclass support changes look good! And it'll be good to have the additional inference parsing test as well. The thing that makes me nervous is the version update just because we've had problems with it before. As far as I can tell it's okay, but maybe we should do some additional checking to make sure? Not sure exactly how much we can do on that front though. 

I'm approving but I think it would be good for Tyler or someone else to get eyes on it too, just in case I missed anything related to the multiclass responses or the version update.

pyproject.toml
honeytung: Yea it is not necessarily to update the SDK version so it is probably safe to just revert back to the previous version?

tyler-romero (approved):
LGTM! Thanks Harry!

PR #236, opened 2025-05-01
Integration test tweaks and temporarily remove model improvement assertion

CoreyEWood:
Temporarily removes the checks for the model being more confident once we fetch an updated binary, since those are currently not reliable because of the OODD implementation. 

This will help the tests pass for now, and later we can revisit how to do this in a more reliable way. We already verify that a new inference pod is created via kubernetes-based checks, so the test will still verify the key behavior of pulling an updated model binary and deploying new inference pods.

Also makes some small organizational tweaks - I left comments in the code about these.


test/api/test_image_queries_live.py
```diff
@@ -60,15 +60,15 @@ def detector(gl: Groundlight) -> Detector:
 def test_post_image_query_via_sdk(gl: Groundlight, detector: Detector):
     """Test that submitting an image query using the edge server proceeds without failure."""
     image_bytes = pil_image_to_bytes(img=Image.open("test/assets/dog.jpeg"))
-    iq = gl.submit_image_query(detector=detector.id, image=image_bytes, wait=10.0)
+    iq = gl.submit_image_query(detector=detector.id, image=image_bytes, wait=10.0, human_review="NEVER")
```
CoreyEWood, line 63: We don't care about the results of these IQs, just that they can get sent through the edge properly. Setting `human_review="NEVER"` ensures that these labels can't skew the performance of the integration test detector at all (since currently this will be run against that detector).

test/assets/cat.jpeg
CoreyEWood, line 1: Standardizing the resolution of this and the dog image.

test/integration/cat.jpg
CoreyEWood, line 1: Switched the integration test to use the `test/assets/[cat]|[dog].jpeg` images for consistency.

test/integration/integration.py
```diff
@@ -8,23 +8,25 @@
...
 import argparse
+import logging
 import os
 import time
 
 import ksuid
 from groundlight import Groundlight
 from model import Detector
 
-EDGE_SETUP = os.getenv("EDGE_SETUP", "0") == "1"
```
CoreyEWood, line 18: This wasn't doing anything helpful and was confusing (the helm test was actually not sending anything to the edge because it didn't set this properly), so I removed it and made it so the integration tests always run against the edge.

test/integration/integration.py
```diff
@@ -8,23 +8,25 @@
...
-ENDPOINT_PORT = os.getenv("EDGE_ENDPOINT_PORT", "30107")
+LOG_LEVEL = os.environ.get("LOG_LEVEL", "INFO").upper()
+logging.basicConfig(
+    level=LOG_LEVEL, format="%(asctime)s.%(msecs)03d %(levelname)s %(message)s", datefmt="%Y-%m-%d %H:%M:%S"
+)
+logger = logging.getLogger(__name__)
```
CoreyEWood, line 23: Using the logger here so that we get the timestamps, for easier debugging.

test/integration/integration.py
```diff
@@ -142,23 +160,5 @@ def _submit_dog_or_cat(detector: Detector, confidence_threshold: float, img_file
...
-            if status_response.status_code == 200:
-                return
-        except requests.RequestException as e:
-            final_exception = e
-            time.sleep(1)  # wait for 1 second before retrying
-    raise Exception(
-        f"Edge endpoint status page is not available after polling for {MAX_WAIT_TIME_S} seconds. {final_exception=}"
-    )
-
-
```
CoreyEWood, line 162: Francine confirmed that this was accidentally duplicated here - it actually lives in `test/integration/status.py`.

test/integration/test-with-k3s-helm.sh
```diff
@@ -124,6 +122,7 @@ helm test -n default ${HELM_RELEASE_NAME} --hide-notes
 
 echo "Helm tests completed successfully."
 
+export EDGE_SETUP=1 # Setting this to 1 will make the integration tests use the edge endpoint
```
CoreyEWood, line 125: Until now, the helm version of the integration tests were actually never sending any requests to the edge.

paulina-positronix (approved):
LGTM as a temporary fix of broken tests. Thanks for the logging!

test/integration/test-with-k3s-helm.sh
tomfaulhaber: Arghh! Good catch!

tomfaulhaber (approved):
Looks great. Thanks!

PR #237, opened 2025-05-01
Make a separate nginx container

tomfaulhaber:
In this PR, I added a new container for nginx. I also left nginx in the old container with a dummy config file that makes it a no-op. This is to make it so folks can pull the chart and containers separately (especially important today because things aren't really deploying right cause of the broken tests). I'll rip that out in a follow-up once everything is fully deployed.


f-wright (approved):
LGTM!

PR #238, opened 2025-05-02
Fix model updating duplication

f-wright:
There was a small bug that was causing us to update each inference deployment twice when we were updating models. This PR switches our updating logic to iterate through detector ids instead of deployment records, so each inference deployment is only updated once.


CoreyEWood (approved):
Nice, looks great!

PR #242, opened 2025-05-07
Check for network connection proactively

CoreyEWood:
Note: this modifies part of the critical path for the edge so please review carefully (and feel free to demand more rigor) / test out yourself if you have the time!

Adds some utility functions to proactively check for network connection. Introduces network connection checks before calling SDK methods via `safe_call_sdk` and before fetching model info (let me know if you think of other places where we should do this too).

This improves behavior while the edge is offline. Instead of attempting to make the request to the cloud and retrying for 10 seconds, the edge will now internally proactively check if it's online before making calls to the cloud. If it isn't, it doesn't attempt to make the request and a 503 is raised. 

If a client submits an IQ to the edge which needs to be escalated to the cloud and the edge is offline, the client will now see
```
groundlight_openapi_client.exceptions.ServiceException: (503)
Reason: Service Unavailable
HTTP response headers: HTTPHeaderDict({'Server': 'nginx/1.18.0', 'Date': 'Wed, 07 May 2025 22:22:53 GMT', 'Content-Type': 'application/json', 'Content-Length': '45', 'Connection': 'keep-alive'})
HTTP response body: {"detail":"No network connection available."}
```
as part of the error they get back.


app/core/utils.py
```diff
@@ -136,20 +147,75 @@ def _mode_to_result_and_type(
...
+
+@cached(connection_ttl_cache)
+def is_connected() -> bool:
+    """Check if the system can establish a network connection to the test host. Result is cached for a short time."""
+    try:
+        socket.setdefaulttimeout(CONNECTION_TIMEOUT)
+        socket.socket(socket.AF_INET, socket.SOCK_STREAM).connect((CONNECTION_TEST_HOST, CONNECTION_TEST_PORT))
+        return True
+    except socket.error:
+        return False
```
CoreyEWood, line 161: This was written with help from ChatGPT, but it seems reliable from my testing. Definitely open to suggestions if there's a better way to do this.

app/core/utils.py
```diff
@@ -136,20 +147,75 @@ def _mode_to_result_and_type(
...
+        try:
+            return api_method(**kwargs)
+        # except groundlight.NotFoundError as ex:
+        #     raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(ex))
+        except Exception as ex:
+            if hasattr(ex, "status"):
+                raise HTTPException(status_code=ex.status, detail=str(ex)) from ex
+            raise ex
+
+    raise HTTPException(status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail="No network connection available.")
```
CoreyEWood, line 218: This ensures that we don't bother trying to make the SDK call if we don't have network connection, and reduces the time we wait/retry. Before this, when there was no network connection it would retry the SDK call for 10 seconds.

app/core/constants.py
```diff
@@ -1 +1,9 @@
 DEFAULT_PATIENCE_TIME: int = 30  # Default patience time in seconds
+
+CONNECTION_STATUS_TTL_SECS = 0.5  # Number of seconds to cache the connection status
```
CoreyEWood, line 3: I'm not sure if we want to cache the connection status or not. In principal it sounds like a good idea, but caching for x seconds interacts kind of strangely with polling for connection with exponential backoff for y seconds. Plus, I don't expect us to make too many repeated checks and each check should be fast.

app/core/utils.py
```diff
@@ -136,20 +147,75 @@ def _mode_to_result_and_type(
...
```
CoreyEWood, line 218: This ensures that we don't bother trying to make the SDK call if we don't have network connection, and reduces the time we wait/retry. Before this, when there was no network connection it would retry the SDK call for 10 seconds (and then return an ugly error message).

app/core/edge_inference.py
```diff
@@ -318,6 +318,11 @@ def update_models_if_available(self, detector_id: str) -> bool:
 
         Returns True if a new model was downloaded and saved, False otherwise.
         """
+        has_connection = wait_for_network_connection(1.0)
```
tyler-romero, line 321: nit: why do we have a default value if we're just setting a magic number to override it?

tyler-romero:
Started to review but I'm still trying to figure out how I feel about this change. Will get back to you.

app/core/edge_inference.py
CoreyEWood: I think originally I was planning on using it with the default value sometimes, but currently there's no instances of that so it probably makes sense to get rid of the default value.

tyler-romero:
I think my main question is: do you think this should just be upstreamed to the SDK?

app/core/utils.py
```diff
@@ -136,20 +147,75 @@ def _mode_to_result_and_type(
...
 
 
+connection_ttl_cache = TTLCache(maxsize=1, ttl=CONNECTION_STATUS_TTL_SECS)
+
+
+@cached(connection_ttl_cache)
+def is_connected() -> bool:
+    """Check if the system can establish a network connection to the test host. Result is cached for a short time."""
+    try:
+        socket.setdefaulttimeout(SOCKET_TIMEOUT)
```
tomfaulhaber, line 157: This will set the timeout globally for the process. Is that what you want? I think I'd just set a specific timeout value for the `socket.socket(...)` call.

app/core/utils.py
```diff
@@ -136,20 +147,75 @@ def _mode_to_result_and_type(
...
+    """Check if the system can establish a network connection to the test host. Result is cached for a short time."""
+    try:
+        socket.setdefaulttimeout(CONNECTION_TIMEOUT)
+        socket.socket(socket.AF_INET, socket.SOCK_STREAM).connect((CONNECTION_TEST_HOST, CONNECTION_TEST_PORT))
+        return True
+    except socket.error:
+        return False
+
+
+def wait_with_exponential_backoff(
```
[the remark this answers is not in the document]
tomfaulhaber: Why do we want exponential backoff here? If you're going to use it and have it run forever, add a maximum time between polls (something like 10s is probably fine). Otherwise, after a while offline, you'll have multi-hour waits between retries.  (That doesn't really matter if you're only waiting 1s, but the default arg is infinity)

app/core/utils.py
```diff
@@ -136,20 +147,75 @@ def _mode_to_result_and_type(
...
 
+connection_ttl_cache = TTLCache(maxsize=1, ttl=CONNECTION_STATUS_TTL_SECS)
+
+
+@cached(connection_ttl_cache)
+def is_connected() -> bool:
+    """Check if the system can establish a network connection to the test host. Result is cached for a short time."""
+    try:
+        socket.setdefaulttimeout(SOCKET_TIMEOUT)
+        socket.socket(socket.AF_INET, socket.SOCK_STREAM).connect((CONNECTION_TEST_HOST, CONNECTION_TEST_PORT))
```
tomfaulhaber, line 158: You need to be sure that you close the socket here otherwise you'll run out of file descriptors

app/core/utils.py
```diff
@@ -136,20 +147,75 @@ def _mode_to_result_and_type(
...
     """
-    try:
-        return api_method(**kwargs)
-    # except groundlight.NotFoundError as ex:
-    #     raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(ex))
-    except Exception as ex:
-        if hasattr(ex, "status"):
-            raise HTTPException(status_code=ex.status, detail=str(ex)) from ex
-        raise ex
+    has_connection = wait_for_network_connection(timeout_sec=1.0)  # Wait for connection for up to one second
```
tomfaulhaber, line 207: Doing this on every call? That seems like a bunch of extra overhead (plus at scale could get us blocked by google.)

app/core/utils.py
```diff
@@ -136,20 +147,75 @@ def _mode_to_result_and_type(
...
+    if has_connection:
+        try:
+            return api_method(**kwargs)
+        # except groundlight.NotFoundError as ex:
```
tomfaulhaber, line 211: ??? Why are we submitting commented out code?

CoreyEWood: If you look at the diff you'll see that this commented out code was previously there (and has been there since October). We can take it out, but it's not related to this PR.

tomfaulhaber: Oh, yeah. I see.

Let's take it out. Commented code is a code smell that we should (a) usually avoid to begin with and (b) should clean up when we get a chance.

Exceptions sometimes for debugging code.

PR #250, opened 2025-06-06
Adjust MINIMUM_EXPECTED_BINARY_QUERY_RATE for edge canary

timmarkhuff:
The laptop edge canary is sometimes narrowly missing the target query rate, so we will lower the target a bit to avoid unnecessary alerts.

### DIFF
--- a/cicd/edge-canaries/laptop/run.py
+++ b/cicd/edge-canaries/laptop/run.py
@@ -97,11 +97,11 @@ logger.info(
 
 # Check that the query rate is sufficiently fast (edge speed)
 # The threshold here is a somewhat arbitrary value that represents a reasonable value for number of queries per second.
-MINIMUM_EXPECTED_BINARY_QUERY_RATE = 10.0 
+MINIMUM_EXPECTED_BINARY_QUERY_RATE = 7.0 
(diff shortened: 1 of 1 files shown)


cicd/edge-canaries/laptop/run.py
```diff
@@ -97,11 +97,11 @@
 
 # Check that the query rate is sufficiently fast (edge speed)
 # The threshold here is a somewhat arbitrary value that represents a reasonable value for number of queries per second.
-MINIMUM_EXPECTED_BINARY_QUERY_RATE = 10.0 
+MINIMUM_EXPECTED_BINARY_QUERY_RATE = 7.0 
 if query_rate < MINIMUM_EXPECTED_BINARY_QUERY_RATE:
     logger.error(
         f"Edge Canary actual binary query rate is {query_rate}, less than expected minimum of {MINIMUM_EXPECTED_BINARY_QUERY_RATE}. "
-        f"There might be something wrong with the Edge Endpoint, or your detector ({detector_id}) might need more labels to function properly."
```
timmarkhuff, line 104: This is no longer an issue of how well the detector is trained, because the detector is now configured with `edge_answers_with_escalation`

PR #251, opened 2025-06-09
Make OODD requests sequential

f-wright:
After benchmarking, I determined that it's actually faster to run the primary and OODD inference requests sequentially, as opposed to in parallel with asyncio. OODD runs pretty quickly, so it seems that the setup time to gather the async results is more than what it takes to just run OODD after primary inference. This PR makes that change, to improve the edge-endpoint's framerate capability.

### DIFF
--- a/app/core/edge_inference.py
+++ b/app/core/edge_inference.py
@@ -1,4 +1,3 @@
-import asyncio
 import logging
 import os
 import shutil
--- a/app/core/edge_inference.py
+++ b/app/core/edge_inference.py
@@ -21,7 +21,7 @@ logger = logging.getLogger(__name__)
 
 # Simple TTL cache for is_edge_inference_ready checks to avoid having to re-check every time a request is processed.
 # This will be process-specific, so each edge-endpoint worker will have its own cache instance.
-ttl_cache = TTLCache(maxsize=128, ttl=2)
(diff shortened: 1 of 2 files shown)


app/core/edge_inference.py
```diff
@@ -21,7 +21,7 @@
 
 # Simple TTL cache for is_edge_inference_ready checks to avoid having to re-check every time a request is processed.
 # This will be process-specific, so each edge-endpoint worker will have its own cache instance.
-ttl_cache = TTLCache(maxsize=128, ttl=2)
+ttl_cache = TTLCache(maxsize=128, ttl=5)
```
f-wright, line 24: Unrelated change, to make our cache stick around a little longer.

paulina-positronix (approved):
LGTM thank you! Will this always be true that async is slower (edge, edge-on-cloud, various configs)?

CoreyEWood (approved):
Looks good 👍

f-wright:
> LGTM thank you! Will this always be true that async is slower (edge, edge-on-cloud, various configs)?

@paulina-positronix I've only tested with a g4 EoC instance, but I don't anticipate that it'd be different elsewhere. Neither the set up work nor the OODD inference are GPU constrained, so I don't think they'd run particularly differently in other edge configurations. Happy to find other machines to test it on though if you think we should do that before merging!

PR #255, opened 2025-06-17
Use groundlight 0.23.0

CoreyEWood:
Updates the groundlight SDK version.

I used `sdk_version_testing.py` to test the previous 30 SDK versions against this (back to 0.15.1) and they were all able to successfully submit an image query.


brandon-wada (approved):
LGTM

PR #256, opened 2025-06-19
Add Kubernetes memory limits to avoid OOM crashes

timmarkhuff:
This PR prevents the OOM system crashes we have been seeing when a user attempts to use too many detectors at the same time.

A Kubernetes eviction policy is added that will start evicting pods when the system-wide available memory drops below a certain threshold. This threshold is defined as either a percentage of available memory, or an absolute number, whichever is smaller.

Additionally, memory "requests" were added for system-critical containers (each of the four containers in the edge-endpoint pod) so that they will not be evicted, which would cause the system to break.

I included `memory_pressure_test.py` to test these changes. The changes are effective; I have never seen the Edge Endpoint crash while these changes are in effect. However, it does occasionally slow down. Also, since it evicts inference pods, the user may not get edge answers for all of their detectors. I think this is acceptable, since these changes are meant to be a fail safe that simply prevents the system from crashing. We can add more sophisticated logic in the future that intelligently loads and offloads inference pods based on usage, user-defined settings, etc.


deploy/helm/groundlight-edge-endpoint/templates/edge-deployment.yaml
```diff
@@ -63,6 +63,9 @@ spec:
       - name: nginx
         image: *edgeEndpointImage
         imagePullPolicy: "{{ include "groundlight-edge-endpoint.edgeEndpointPullPolicy" . }}"
+        resources:
+          requests:
+            memory: "50Mi"
```
robotrapta, line 68: Your PR comment implies that these requests protect these pods from eviction.  Could you add a comment to that effect in the YAML here?

load-testing/memory-pressure-test/README.md
```diff
@@ -0,0 +1,10 @@
+# Memory Pressure Test
+Tools that test the Edge Endpoint's resilience under memory pressure by spawning multiple inference pods simultaneously.
+
+## To Run
+1. Create an environment: `python3 -m venv .venv`
```
robotrapta, line 5: You should start learning `uv` so that it's your go-to default for things like this.  uv is the future.

load-testing/memory-pressure-test/README.md
robotrapta, line 1: Nice README!

deploy/helm/groundlight-edge-endpoint/values.yaml
```diff
@@ -55,4 +55,17 @@ logLevel: "INFO"
...
\ No newline at end of file
+fullnameOverride: "edge-endpoint"
+
+# Set some sensible limits on memory usage to avoid system crashing
+k3sConfig:
+  enabled: true
+  evictionHardPercent: "15"
+  evictionSoftPercent: "25" 
+  evictionHardMinGB: "4"
+  evictionSoftMinGB: "8"
```
robotrapta, line 66: These really seem high to me.  Have you tested these on a jetson?

deploy/helm/groundlight-edge-endpoint/templates/k3s-memory-config.yaml
```diff
@@ -0,0 +1,107 @@
...
+      serviceAccountName: {{ include "groundlight-edge-endpoint.serviceAccountName" . }}-k3s-config
+      hostNetwork: true
+      hostPID: true
+      containers:
+      - name: k3s-memory-configurator
+        image: alpine:latest
+        command:
+        - /bin/sh
+        - -c
+        - |
```
robotrapta, line 24: I wish there was a simple way to put this .sh script into its own file and get the same results.  This is fine, but it's a bit ugly - IDE's dont understand it - and kinda hard to test.  And I don't have a simple answer for how you do this without building another container, which is a big jump up in complexity.

robotrapta (approved):
This looks good to me.  Biggest question is about the default eviction minimum values - 4GB seems very conservative especially for e.g. an 8GB Jetson.  Have we tried this on an Orin?

timmarkhuff:
> This looks good to me. Biggest question is about the default eviction minimum values - 4GB seems very conservative especially for e.g. an 8GB Jetson. Have we tried this on an Orin?

I haven't tried it on an Orin, but I can.

The way the memory limits work is that it will use either a specified percentage of the total system memory, or a fixed amount, whichever is smaller. This is to avoid using too much memory on low-memory machines, while at the same time not being overly liberal with memory allocation on high-memory machines. Does this seem reasonable?

deploy/helm/groundlight-edge-endpoint/templates/k3s-memory-config.yaml
```diff
@@ -0,0 +1,107 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "groundlight-edge-endpoint.fullname" . }}-k3s-memory-config
+  namespace: {{ .Values.namespace }}
+  labels:
+    {{- include "groundlight-edge-endpoint.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
```
tomfaulhaber, line 9: This will replace the config file on every upgrade. Is that what we want? (I guess for existing systems we need that).

deploy/helm/groundlight-edge-endpoint/templates/k3s-memory-config.yaml
```diff
@@ -0,0 +1,107 @@
...
+            SOFT_MIN_GI=$((SOFT_MIN_GB * 1000000000 / 1073741824))
+            SOFT_THRESHOLD="${SOFT_MIN_GI}Gi"
+            echo "Using soft threshold: ${SOFT_MIN_GB}GB (${SOFT_THRESHOLD}) - absolute minimum"
+          else
+            SOFT_THRESHOLD="${SOFT_PERCENT}%"
+            echo "Using soft threshold: ${SOFT_THRESHOLD} (~${SOFT_PERCENT_GB}GB)"
+          fi
+          
+          mkdir -p /etc/rancher/k3s
+          cat > /etc/rancher/k3s/config.yaml << EOF
```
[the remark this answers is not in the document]
timmarkhuff: I factored out the shell script into a separate file and included it in the job like this: `{{- tpl (.Files.Get "files/k3s-memory-config.sh") . | nindent 10 }}`

Seems to work well. Thank you for the suggestion.

deploy/helm/groundlight-edge-endpoint/templates/edge-deployment.yaml
timmarkhuff: Done.

load-testing/memory-pressure-test/README.md
timmarkhuff: Done. The future is beautiful.

deploy/helm/groundlight-edge-endpoint/templates/k3s-memory-config.yaml
timmarkhuff: I factored it out into a separate script per Tom's recommendation. Works great!

honeytung (approved):
Had a discussion with Tim on this. I believe Tim also implemented a memory percentage limit instead of hard numerical limit on smaller devices such as Jetsons. If this change is working on those devices then I think it is good to go. Thanks for the change!

PR #258, opened 2025-06-25
load-testing README updated to use helm commands

diane-park:
- Added helm commands to install edge-endpoint with new configurations: 
- Has all the steps to work for running in a different server than where the endpoint (resetting env variable)

### DIFF
--- a/load-testing/README.md
+++ b/load-testing/README.md
@@ -29,7 +29,35 @@ command:
 
 Some trial and error will likely be necessary to figure out the ideal configuration.
 
-After setting these config options, you should run/re-run the [setup edge endpoint script](/deploy/bin/setup-ee.sh) to deploy with your new configuration. You can monitor the inference pod's logs to see when all of the workers have finished starting up (if the number of workers is high, this will likely be after the pod reports being ready). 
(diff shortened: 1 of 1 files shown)


load-testing/README.md
```diff
@@ -29,7 +29,35 @@ command:
 
 Some trial and error will likely be necessary to figure out the ideal configuration.
 
-After setting these config options, you should run/re-run the [setup edge endpoint script](/deploy/bin/setup-ee.sh) to deploy with your new configuration. You can monitor the inference pod's logs to see when all of the workers have finished starting up (if the number of workers is high, this will likely be after the pod reports being ready). 
+After setting these config options, you should run/re-run the edge-endpoint installation command to deploy with your new configuration. 
+
+#### If using setup-ee.hm:
```
f-wright, line 34:
```suggestion
#### If using setup-ee.sh: 
```

f-wright (commented):
Thanks for updating the readme with your fix!!

Left a couple of comments -- mostly wondering about the additional helm commands, before the upgrade. I think just running the upgrade should be ok?

Also, might be good to move the helm section above the setup-ee section, since we mostly use helm now.

PR #259, opened 2025-06-30
Fixing edge metrics error message

timmarkhuff:
There is an error message that always appears at the top of the edge metrics page, even if there is no error. This is caused by an incorrect ID in the html. 

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,7 @@ RUN apt-get update && \
     sqlite3 && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/* && \
-    curl -sSL https://install.python-poetry.org | python - && \
+    POETRY_HOME=${POETRY_HOME} curl -sSL https://install.python-poetry.org | python - && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -36,7 +36,7 @@ RUN apt-get update && \
     sqlite3 && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/* && \
-    curl -sSL https://install.python-poetry.org | python - && \
+    POETRY_HOME=${POETRY_HOME} curl -sSL https://install.python-poetry.org | python - && \
```
timmarkhuff, line 39: I couldn't get this script to successfully run without making this change. I'm curious if anyone else has had this problem.

f-wright: What were you trying to run? I don't think I've run into this issue

timmarkhuff: Sorry, I left out that key bit of context :) 

I was trying to run `/deploy/bin/build-local-edge-endpoint-image.sh`

PR #263, opened 2025-07-02
Fixing k3s memory script

timmarkhuff:
There was a bug in this script such that if you ran it a second time, it would incorrectly edit `kubelet-args` and leave them in a broken state. This PR fixes the issue.

This script now correctly handles the following situations:
-if the config file does not yet exist, it will be created with the correct kubelet-args
-if the config file exists, but there are no kubelet-args, our kubelet-args will be appended and the other content in the file will be untouched
-if the config file exists, and there are already kubelet-args defined (other than the one this script defines), our kubelet-args are appended to the kubelet-args section, leaving other user-defined configurations intact


CoreyEWood (approved):
This looks reasonable to me, though I don't understand perfectly everything that's happening. But if you tested it and it worked, that seems good enough.

PR #264, opened 2025-07-10
add a little bit more explaination in our example config

brandon-wada:


configs/edge-config.yaml
```diff
@@ -27,5 +27,7 @@ edge_inference_configs: # These configs define detector-specific behavior and ca
 detectors: # Each entry here defines the edge configuration for a detector.
   # To configure a detector, add an entry for it below with its detector ID and the name of the edge inference config
   # you want to use. You can use or modify one of the existing edge inference configs listed above, or define your own.
-    - detector_id: ""
```
brandon-wada, line 30: Would like a sanity check that there wasn't something special about the empty string id here. Will this change cause the edge to attempt to spin up two inference pods for the example here?

brandon-wada:
This all is actually written up better in CONFIGURING-DETECTORS.md, we should just defer to there

PR #266, opened 2025-07-10
Enable running in minimal mode

f-wright:
This PR enables running the edge endpoint in "minimal mode", which should use less memory and run faster. Currently, minimal mode has limited capabilities (only binary and multiclass detectors available), so it is not enabled as the default for all deployments yet. 

We store our minimal images in a separate ECR repository from our standard pipeline images. This PR adds an option `useMinimalImage` to helm, which will pull the image from our minimal repository if enabled. 

This PR also prevents model saving, pod launching, and inference for a separate OODD model if `useMinimalImage` is enabled.


tomfaulhaber (approved):
LGTM!

I'll be glad when we always use the minimal model container and we can clean up all the conditionals in this code.

PR #268, opened 2025-07-24
Making Edge-Only work on Balena

timmarkhuff:
Changes:
* now works with Balena
* now totally independent of the cloud service (models are stored in s3 and fetched from there)

### DIFF
--- a/app/core/edge_inference.py
+++ b/app/core/edge_inference.py
@@ -479,6 +479,7 @@ def should_update(model_dir: str, version: Optional[int]) -> bool:
     else:
         return False
 
+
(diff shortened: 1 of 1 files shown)


deploy/k3s/inference_deployment/inference_deployment_template.yaml
```diff
@@ -65,9 +63,29 @@ spec:
         volumeMounts:
         - name: pina-models
           mountPath: /opt/models
+      - name: sync-edge-endpoint-state
```
timmarkhuff, line 66: Syncing all components of the edge endpoint state (models etc.) that would normally be pulled down from the service

PR #270, opened 2025-08-11
Distributing inference pods across available GPUs

timmarkhuff:
This PR allows machines with multiple GPUs to utilize all of their GPUs.

I have updated the inference image to contain a script for selecting a GPU, if there are multiple GPUs. 

I also refactored the memory pressure script so that it could be used for the sort of testing I was doing on this PR.

### DIFF
--- a/load-testing/README.md
+++ b/load-testing/README.md
@@ -14,18 +14,23 @@ detectors:
     edge_inference_config: "no_cloud"
 ```
 
-* You'll want to configure the edge-endpoint proxy and the inference server to have the optimal number of workers. A general guideline is that your number of edge-endpoint proxy workers shouldn't exceed the number of CPU cores on your machine. Try to set your number of inference workers to maximize GPU utilization (if using a GPU).
-    * To increase the number of edge-endpoint proxy workers, change the `--workers` param in the CMD line of the [Dockerfile](/Dockerfile). 
-    * To increase the number of inference server workers, in the [inference_deployment_template](/deploy/k3s/inference_deployment/inference_deployment_template.yaml) locate the below command and change the argument for `--workers`.
-```
-command:
-    [
-        "poetry", "run", "python3", "-m", "uvicorn", "serving.edge_inference_server.fastapi_server:app",
-        "--host", "0.0.0.0",
-        "--port", "8000",
-        "--workers", "1"
-    ]
-```
+* You'll want to configure both the edge endpoint and the inference server to have the optimal number of workers. A general guideline is that your number of edge endpoint workers shouldn't exceed the number of CPU cores on your machine. Try to set your number of inference workers to maximize GPU utilization (if using a GPU).
+
+    * **To configure edge endpoint workers**: Edit the `--workers` parameter in [launch-edge-logic-server.sh](/app/bin/launch-edge-logic-server.sh). Change from the default:
+    ```bash
+    poetry run uvicorn \
+        --workers 8 \  # You can tweak this for load testing
+        --host 0.0.0.0 \
+        --port ${APP_PORT} \
+        --proxy-headers \
+        app.main:app
+    ```
+
+    * **To configure inference server workers**: Set the `WORKERS` environment variable in the inference deployment template at [inference-deployment-template.yaml](/deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml). Change the value from `"1"` to your desired number of workers:
+    ```yaml
+    - name: WORKERS  # Number of uvicorn workers for the inference server
+      value: "4"  # Increase this number for load testing
+    ```
 
 Some trial and error will likely be necessary to figure out the ideal configuration.
 


deploy/helm/groundlight-edge-endpoint/files/select-gpu.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
timmarkhuff, line 1: Not sure if we should keep this script. On the one hand, it seems like it might be useful, but on the other hand, my testing showed that it was not as good as `randomly-select-gpu.sh`, so we currently aren't using this script.

deploy/helm/groundlight-edge-endpoint/files/randomly-select-gpu.sh
```diff
@@ -0,0 +1,28 @@
...
+    SELECTED_GPU=$((RANDOM % GPU_COUNT))
+    echo "Multiple GPUs detected, randomly selected GPU $SELECTED_GPU out of $GPU_COUNT" >&2
+    echo "$SELECTED_GPU"  # Output the selected GPU ID
+  else
+    echo "No GPUs detected, running on CPU" >&2
+    echo "Running on CPU (no GPU selected)" >&2
+    echo ""  # Output empty string for CPU mode
+  fi
+else
+  echo "nvidia-smi not available, running on CPU" >&2
```
robotrapta, line 25: This feels like the wrong error logic.  For a production system like this, we'd prefer it to fail loudly with a clear error message vs limp along at reduced capacity.  The latter leads to disappointed customers not understanding why the system is performing badly.  General rule for production systems is to either work perfectly or fail loudly.

Ideally we can tell (from helm) if the system is configured to use GPU or not.  If it is, then we should insist on `nvidia-smi` being available, and if it's not, then we error out.  (This script probably should have `set -e` at the top so if any command fails, then the script fails.)

deploy/helm/groundlight-edge-endpoint/files/select-gpu.sh
robotrapta: I think it's reasonable to leave here, but your findings about it not working well (and why!) should be in a comment.  This helps somebody reach the same conclusion faster.

Also, I think we should add a second option before too long.  "Load balanced" I think would be a good one - where it assigns to the GPU with the fewest number of processes running on it.  And randomly selects between GPUs with equal number of processes.  This should work better in a world where the GPUs are compute limited more than VRAM limited.  But that can be a second PR.  Also, at that point I get the feeling we should move the GPU selection logic into python.

load-testing/memory-pressure-test/memory_pressure_test.py
```diff
@@ -4,6 +4,7 @@
 import os
```
robotrapta, line 4: Is this script described in a README somewhere?  Either there or here at the top, we should have a good paragraph explaining how you run it.

robotrapta (approved):
One important change requested and a few suggestions to improve.

I think we should do a follow-up to have configurable strategies for GPU load balancing.

load-testing/memory-pressure-test/memory_pressure_test.py
timmarkhuff: There was a README. I just updated it to make it more helpful.

deploy/helm/groundlight-edge-endpoint/files/randomly-select-gpu.sh
timmarkhuff: I addressed this and moved the script into zuuul so it can live with the inference container

deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
```diff
@@ -75,13 +75,10 @@ spec:
           value: {{ .Values.logLevel | quote }}
         - name: LOAD_ALL_PIPELINES  # Load only the pipelines that are needed for edge inference.
           value: "false"
-        command:
```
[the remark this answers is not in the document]
timmarkhuff: I have tested it end to end and it works.

`launch-inference-server.sh` is in zuuul and it is defined as the entry point for the container.

Previously there was a `launch-inference-server.sh` defined here in `edge-endpoint` but it was almost exactly a duplicate of the one on zuuul; the only functionality it added was the ability to pass in a different number of workers, which was useful for performance testing. I updated the script in zuuul to support workers as an argument, so we can still tinker with that as needed.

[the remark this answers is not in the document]
robotrapta: Yup - I found it now.  Thanks!

deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
CoreyEWood, line 1: I think this file should be backwards compatible with the old inference image, since that already calls the launch script itself and won't be adversely affected by the new env vars set here.

CoreyEWood (approved):
Looks good!

load-testing/memory-pressure-test/memory_pressure_test.py
```diff
@@ -94,142 +98,165 @@ def pprint_iq(iq: ImageQuery, confidence_threshold: float) -> None:
...
+        print(f"{n + 1}: Got or created {detector.id}")
+        detectors.append(detector)
+    return detectors
 
-            if not iq.result.from_edge:
-                add_label_async(gl, iq, label)
-                print(f'Added {label} label to {iq.id} on {d.id}.')
-                
-def main_count(num_detectors: int) -> None:
+def get_or_create_count_detectors(
```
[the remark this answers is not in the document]
timmarkhuff: I agree!

load-testing/memory-pressure-test/memory_pressure_test.py
```diff
@@ -21,19 +22,36 @@
...
+            return gl
+        except Exception as e:
+            now = time.time()
+            elapsed_time = now - start_time
+            if elapsed_time > timeout_sec:
+                raise RuntimeError(
+                    f'Unable to connect to Groundlight within timeout of {timeout_sec} seconds: {e}'
+                )
+            print(f'Failed to connect to Groundlight after {int(elapsed_time)} seconds: {type(e).__name__}. Retrying...')
+            time.sleep(1)
```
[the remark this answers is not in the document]
timmarkhuff: I attempted to explain this in the readme for this script, but I recognize it's still a little confusing.

PR #276, opened 2025-09-04
In setup-ee, use the edge-endpoint container for sync_pinamod

tomfaulhaber:
This keeps us from using public containers from docker.

I haven't thought about how to do this from helm yet where we have a chicken or egg problem.

Also, I fixed the edge-endpoint container build to be platform aware when installing `aws-cli`.


f-wright (approved):
LGTM

PR #280, opened 2025-09-09
Don't erase container images in the K3s pre-load directory

tomfaulhaber:
This lets us load images onto machines so they don't need to be pulled from Docker Hub. More automation coming.


f-wright (approved):
LGTM!

honeytung (approved):
LGTM!

PR #286, opened 2025-09-15
Include submitted human review value when writing an unconfident query to the queue

CoreyEWood:
Previously this was mistakenly always submitting with `None`, which didn't respect the user's inputted value.


timmarkhuff (approved):
LGTM

PR #290, opened 2025-09-18
Improvements to queue escalation error handling and retries

CoreyEWood:
Changes made in response to manual network fault testing I've done with Toxiproxy and Chaos Mesh.
- Catches `ReadTimeoutError` when escalating from the queue and prompts a retry, since we expect that this exception means we have bad network connectivity.
- Sets different timeout values for connect vs read and increases the read timeout period to avoid failing in bad network connectivity settings when a longer read timeout value would have allowed the request to succeed.
  - These values are still fairly arbitrary (based on manual testing) so we should keep evaluating and tweaking these values to make sure they achieve the behavior we want.
- Improves logging around retries and caught exceptions.

### DIFF
--- a/app/escalation_queue/manage_reader.py
+++ b/app/escalation_queue/manage_reader.py
@@ -7,7 +7,7 @@ from pathlib import Path
 
 from fastapi import HTTPException, status
 from groundlight import Groundlight, GroundlightClientError, ImageQuery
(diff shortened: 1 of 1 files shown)


tomfaulhaber (approved):
Nice!

PR #292, opened 2025-09-29
Removing OpenCV dependency

timmarkhuff:
The poetry environment for the `edge-endpoint` container contains OpenCV, which actually isn't used by the container. In order to remove cruft and make the image smaller, I have removed OpenCV and its system-level dependencies.

Doing this reduced the docker image size from 1.29GB to 1.14GB.

Additionally, I decoupled the `load-testing` scripts from the `edge-endpoint`'s environment, giving them their own `uv` environment. I also moved around and cleaned up some of the scripts I added and their corresponding documentation.


CoreyEWood (approved):
Nice, removing an unnecessary package is a big win! 

Organization/documentation improvements for the load testing tools also look great.

PR #293, opened 2025-09-30
Fixing failing CICD

timmarkhuff:
Fixing a failing CICD test: https://github.com/groundlight/edge-endpoint/actions/runs/18144199314/job/51648598272


Makefile
```diff
@@ -55,7 +55,7 @@ helm-install:
 	helm upgrade -i -n default --dependency-update ${HELM_ARGS} ${HELM_RELEASE_NAME} deploy/helm/groundlight-edge-endpoint 
 
 helm-package:
-	helm package deploy/helm/groundlight-edge-endpoint
```
timmarkhuff, line 58: This is unrelated to the failing CICD, but probably still a good idea

Makefile
```diff
@@ -55,7 +55,7 @@ helm-install:
...
+	helm package deploy/helm/groundlight-edge-endpoint --dependency-update
```
timmarkhuff, line 58: This is unrelated to the failing CICD, but probably still a good idea

PR #353, opened 2026-03-13
New`/status/resources.json` endpoint and RAM metrics on "Groundlight Edge Endpoint Status" page

timmarkhuff:
## Changes
1. `/status/gpu.json` is replaced with `/status/resources.json` and reshapes the payload so it is grouped by resource type (`ram`, `vram`) at both the system level and per-detector. This endpoint now provides not only detector-level VRAM metrics, but also RAM. We could add more resources here as needed without rearranging top-level keys. The diff between the old and new response is shown below:
```diff
- {
-     "total_vram_bytes": 16106127360,
-     "used_vram_bytes": 2824011776,
-     "detectors": [
-         {
-             "detector_id": "det_2z41rs0Fo12LAk0oOZg0r4wR9Fn",
-             "primary_vram_bytes": 392167424,
-             "oodd_vram_bytes": 392167424,
-             "total_vram_bytes": 784334848
-         },
-         ...
-     ],
-     "loading_vram_bytes": 0,
-     "observed_gpus": [
-         {
-             "name": "Tesla T4",
-             "total_vram_bytes": 16106127360,
-             "used_vram_bytes": 2824011776,
-             "index": 0
-         }
-     ]
- }
+ {
+     "system": {
+         "ram": {
+             "used_bytes": 11350978560,
+             "total_bytes": 16455831552,
+             "loading_detectors_bytes": 0,
+             "edge_endpoint_bytes": 968884224,
+             "eviction_threshold_pct": 75
+         },
+         "vram": {
+             "used_bytes": 2824011776,
+             "total_bytes": 16106127360,
+             "loading_detectors_bytes": 0,
+             "edge_endpoint_bytes": 0,
+             "observed_gpus": [
+                 { "name": "Tesla T4", "index": 0, "used_bytes": 2824011776, "total_bytes": 16106127360 }
+             ]
+         }
+     },
+     "detectors": [
+         {
+             "detector_id": "det_2z41rs0Fo12LAk0oOZg0r4wR9Fn",
+             "ram":  { "primary_bytes": 818372608, "oodd_bytes": 811892736, "total_bytes": 1630265344 },
+             "vram": { "primary_bytes": 392167424, "oodd_bytes": 392167424, "total_bytes": 784334848 }
+         },
+         ...
+     ]
+ }
```

2. On the "Groundlight Edge Endpoint Status" page, a RAM donut chart was added. The Kubernetes pod eviction threshold is shown on this chart, which will make it easier for users to see if their system is degraded due to memory pressure.
image

3. The React code was also refactored a bit to add the RAM donut chart, fix the chart animations, and generally improve the React code quality.

4. Added an "Edge Endpoint" slice to the RAM donut (black) that accounts for RAM used by all non-detector processes in the Edge Endpoint (the edge-endpoint server, network healer, sidecars, etc.). Surfaced via a new `edge_endpoint_bytes` field on `system.ram` and `system.vram`.

5. Added hover help text for each of the four system legend entries (Loading Detector Models, Edge Endpoint, Other, Free), shown both in the donut slice tooltip and on legend-row hover.

6. Added a `app/status_monitor/dev/` directory with tooling for iterating on the status page frontend without needing a real edge endpoint or Kubernetes cluster. `run-dev.sh` launches the processes.

7. `/status/resources.json` now queries per-pod GPU endpoints in parallel, so worst-case latency is bounded by a single ~2s HTTP timeout regardless of detector count.

8. Kubernetes RBAC updates in `deploy/helm/.../service-account.yaml`:
   - Added `metrics.k8s.io/pods` (list) to the existing namespaced `Role` — needed for per-pod RAM attribution via the Metrics Server.
   - Added a new cluster-scoped `ClusterRole` + `ClusterRoleBinding` granting `nodes` (get, list) and `metrics.k8s.io/nodes` (get, list). Needed to read node capacity and node-level RAM usage. A `ClusterRole` is required because `nodes` is not a namespaced resource.


app/metrics/resource_metrics.py
```diff
@@ -0,0 +1,405 @@
...
+    Uses the node's capacity for total RAM and the Metrics Server for current
+    usage. This matches the kubelet's view of memory, which drives pod eviction.
+    Also extracts the kubelet's soft eviction threshold.
+    """
+    try:
+        nodes = v1.list_node()
+        if not nodes.items:
+            raise ValueError("No nodes found")
+        # The edge endpoint is deployed on single-node k3s; nodes.items[0] is
+        # the only node. On a hypothetical multi-node cluster this would pick
```
f-wright, line 149: Is it worth erroring in this case instead of silently hiding capacity? I don't think we have plans to expand to a multi-node cluster, but this seems like the kind of thing we might forget to update if we did.

f-wright: Claude also suggested `status.nodeName` could be used instead

app/metrics/resource_metrics.py
```diff
@@ -0,0 +1,405 @@
...
+                    available_pct = int(m.group(1))
+                    return 100 - available_pct
+    return None
+
+
+def _find_inference_pods(pod_list) -> list[tuple]:
+    """Return (pod, detector_id, is_oodd, is_ready) for running inference pods."""
+    results = []
+    for pod in pod_list.items:
+        if not pod.status or pod.status.phase != "Running":
```
f-wright, line 211: Should this include pending pods?

app/metrics/resource_metrics.py
```diff
@@ -0,0 +1,405 @@
...
+            detectors, loading_vram, loading_ram = [], 0, 0
+
+        # Anything in the namespace that isn't a recognised Running inference
+        # pod counts as "edge-endpoint platform overhead": the edge-endpoint
+        # server itself, plus sidecars (network-healer, splunk, otel, etc.).
+        # This is a broad heuristic; a stray Job or a user-deployed pod in the
+        # edge namespace would also land in this bucket. Acceptable for now,
+        # since the namespace is chart-owned and unlikely to host foreign pods.
+        inference_pod_names = {pod.metadata.name for pod, _, _, _ in inference_pods}
+        edge_endpoint_ram = sum(
```
f-wright, line 112: I think loading inference pods get lumped into this edge_endpoint_ram total, which seems incorrect.

f-wright: Maybe inference pod names should include all inference pods, regardless of status

app/metrics/resource_metrics.py
```diff
@@ -0,0 +1,405 @@
+"""Collects per-detector GPU and RAM usage for the status page.
```
f-wright, line 1: It doesn't look like there's any test coverage on this work. Is the kubernetes stuff difficult to set up effective tests for? Is there anywhere we can add coverage here?

app/status_monitor/status_web.py
```diff
@@ -46,10 +46,10 @@ async def get_metrics():
     return reporter.metrics_payload()
 
 
-@app.get("/status/gpu.json")
-def get_gpu():
-    """Return per-detector GPU usage as JSON."""
-    return gpu_collector.collect()
+@app.get("/status/resources.json")
+def get_resources():
```
f-wright, line 50: This gets regularly polled, right? It could be cool to display a plot of the resource usage over time when the status page is open. But that's not for this PR, just an idea :)

f-wright (commented):
A couple of small comments, and I'm wondering if we could add test coverage here. I could be convinced to approve without it if you don't think it's worth adding now since this is just for the dashboard, but it seems like this is getting big enough that it might merit more testing.

app/metrics/resource_metrics.py
timmarkhuff: Good suggestion. I implemented this.

app/metrics/resource_metrics.py
timmarkhuff: I think probably not. Pending pods don't have an IP, so it wouldn't be possible to query them for VRAM. Adding pending pods means I have to gate certain parts of the code in order to not query the pending pods for VRAM, which increases complexity.

Claude said there could be some brief situations were pending pods consume RAM, and we might want to include them so that they get attributed to "Loading Detector Models" rather than "Edge Endpoint". I ran a few tests and plotted the results, and could not demonstrate this to be true. 

I think I'd rather not touch this for now.

app/status_monitor/status_web.py
timmarkhuff: Yes, I like that idea!

app/metrics/resource_metrics.py
timmarkhuff: Pending pods are intentionally excluded because they cannot be queried for VRAM (they don't have an IP yet). They also don't have any RAM usage yet, so we aren't missing that either.

I ran some tests, and we don't seem to be misattributing anything during pod rollouts.

app/metrics/resource_metrics.py
timmarkhuff: I have added tests at test/metrics/test_resource_metrics.py

app/metrics/resource_metrics.py
f-wright: Thanks for checking!

f-wright (approved):
LGTM, thanks Tim!

PR #184, opened 2025-02-25
COM-2234: Use Helm to deploy Edge Endpoint

tomfaulhaber:
Features added in this PR:
1. Convert the edge endpoint deployment to use helm rather than a mix of `kustomize` and shell scripts. This supports installation without requiring a `git clone` and cleaner, more flexible customization.
2. Standardize the deployment to use the `edge` namespace by default and explicitly disallow deploying to the `default` namespace.
3. Use temporary AWS credentials supplied by Janzu calls rather than requiring AWS credentials for each customer.
4. Support rational image tags for `edge-endpoint` as described in [Tag released containers by tomfaulhaber · Pull Request #188 · groundlight/edge-endpoint](https://github.com/groundlight/edge-endpoint/pull/188)
5. Support local development for `edge-endpoint` without having to push images to ECR.

Note that 4 & 5 have been implemented for the inference container in zuuul. It follows the same pattern as here.

Right now, we run all the same tests with both the helm install and the old `setup-ee.sh`-based install. This is redundant but feels safe while we're actually using both. We'll remove the old ones when we deprecate `setup-ee.sh`.

#### Follow up items
There's still work to be done here that I will do in future pull requests:
- Support "clone-free installs" ([COM-2300 Support clone-free installs of edge endpoint via a helm repo - Jira](https://positronixcorp.atlassian.net/browse/COM-2300))
- Move GLHub installation to helm install rather than submodule-based install ([COM-2301 Modify GLHub to use the helm chart install of Edge Endpoint rather than having a submodule - Jira](https://positronixcorp.atlassian.net/browse/COM-2301))
- Deprecate non-helm installs ([COM-2304 Deprecate non-helm installs for Edge Endpoint - Jira](https://positronixcorp.atlassian.net/browse/COM-2304))
- Automatically determine the inference flavor at run-time. Remove the option (or make it optional) ([COM-2305 Automatically determine whether to use the GPU rather than forcing user to tell you - Jira](https://positronixcorp.atlassian.net/browse/COM-2305))

And some smaller stuff

PR #368, opened 2026-03-31
Fix Inference Pod Failures on Slow Networks

honeytung:
### Summary

- **Warmup-first coordination**: Added a `wait-for-warmup` init container to inference deployments (both Helm and K3s templates) that waits for the `warmup-inference-model` Job to complete before running `sync-pinamod`. This eliminates a race condition where multiple inference pods concurrently running `aws s3 sync --delete` on a shared hostPath volume would delete each other's temporary files, causing rename failures and crash loops.
- **RBAC update**: Granted `get`, `list`, `watch` on `batch/jobs` to the existing service account so the `kubectl wait` in the init container can function.
- **Startup probe timeout increase**: Bumped `startupProbe.failureThreshold` from 120 (20 min) to 270 (45 min) to accommodate slow `timm`/HuggingFace backbone model downloads on constrained networks.
- **Missing `INFERENCE_FLAVOR` env var**: Added `INFERENCE_FLAVOR=gpu` to the K3s inference deployment template, which was present in the Helm template but missing from K3s, causing inference pods to not utilize the GPU.

### Test plan

- [x] Tested on a G4 EC2 instance with network throttling (~2 MiB/s) to verify warmup completes before inference pods start syncing. The process took around 1 hour to for `warmup-inference-model` to download the inference image (3.7GB) and run `sync-pinamod` (2.6GB)
- [x] Tested working on new inference pod rollouts (after cold boot)
- [x] Verified startup probe does not prematurely kill pods during long timm downloads

### DIFF
--- a/deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
+++ b/deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
@@ -48,20 +48,9 @@ spec:
 
       initContainers:
       - name: wait-for-warmup
-        image: bitnamilegacy/kubectl:latest
-        imagePullPolicy: IfNotPresent
-        command:
-          - sh
-          - -c
-          - |
-            NS=$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)
-            echo "Checking for warmup-inference-model job..."
-            if kubectl get job warmup-inference-model -n $NS &>/dev/null; then
-              echo "Waiting for warmup-inference-model to complete..."
-              kubectl wait --for=condition=complete job/warmup-inference-model -n $NS --timeout=1800s || echo "Wait timed out or job failed, proceeding anyway."
-            else
-              echo "Warmup job not found, proceeding."
-            fi
+        image: 767397850842.dkr.ecr.us-west-2.amazonaws.com/edge-endpoint:{{ include "groundlight-edge-endpoint.edgeEndpointTag" . }}
+        imagePullPolicy: "{{ include "groundlight-edge-endpoint.edgeEndpointPullPolicy" . }}"
+        command: ["/bin/bash", "/groundlight-edge/deploy/bin/wait-for-warmup.sh"]
 
       # NOTE: the sync-pinamod container is duplicated in the warmup_inference_model.yaml Job
       # TODO: refactor to share code between the Job and the initContainer in the Deployment
--- a/deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
+++ b/deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
@@ -37,6 +37,7 @@ spec:
 {{ if eq .Values.inferenceFlavor "gpu" }}
       runtimeClassName: nvidia  # Required for GPU use in k3s
 {{- end }}
+      serviceAccountName: edge-endpoint-service-account
       imagePullSecrets:
       - name: registry-credentials
       strategy:
@@ -46,6 +47,22 @@ spec:
           maxUnavailable: 0  # Aim for no downtime during rollout
 
       initContainers:
+      - name: wait-for-warmup
+        image: bitnamilegacy/kubectl:latest
+        imagePullPolicy: IfNotPresent
+        command:
+          - sh
+          - -c
+          - |
+            NS=$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)
+            echo "Checking for warmup-inference-model job..."
+            if kubectl get job warmup-inference-model -n $NS &>/dev/null; then
+              echo "Waiting for warmup-inference-model to complete..."
+              kubectl wait --for=condition=complete job/warmup-inference-model -n $NS --timeout=1800s || echo "Wait timed out or job failed, proceeding anyway."
+            else
+              echo "Warmup job not found, proceeding."
+            fi
+
       # NOTE: the sync-pinamod container is duplicated in the warmup_inference_model.yaml Job
       # TODO: refactor to share code between the Job and the initContainer in the Deployment
       - name: sync-pinamod
--- a/deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
+++ b/deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
@@ -116,7 +116,7 @@ spec:
             port: 8000
           initialDelaySeconds: 10
           periodSeconds: 10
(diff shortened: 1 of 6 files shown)


deploy/k3s/inference_deployment/inference_deployment_template.yaml
```diff
@@ -33,6 +33,7 @@ spec:
     spec:
```
honeytung, line 33: I think we should keep this as a reference until everyone deprecates `setup-ee` then we can remove this folder.

Note that the Dockerfile copies this template but it is not being used (helm overwrites it), but removing this may affect `setup-ee` when we tagged this image as release.

deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
```diff
@@ -99,7 +116,7 @@ spec:
             port: 8000
           initialDelaySeconds: 10
           periodSeconds: 10
-          failureThreshold: 120  # Wait for up to 20 min
+          failureThreshold: 270  # Wait for up to 45 min
```
honeytung, line 108: We bumped the wait time from 20 to 45mins to give us more time to download weights from HF. We will need to revisit this when we figured out how to cache those too.

deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
```diff
@@ -46,6 +47,22 @@ spec:
...
 
       initContainers:
+      - name: wait-for-warmup
+        image: bitnamilegacy/kubectl:latest
+        imagePullPolicy: IfNotPresent
+        command:
+          - sh
+          - -c
+          - |
+            NS=$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)
```
timmarkhuff: Should this be factored out into a separate shell script for legibility?

deploy/k3s/inference_deployment/inference_deployment_template.yaml
```diff
@@ -42,6 +43,22 @@ spec:
...
 
       initContainers:
+      - name: wait-for-warmup
+        image: bitnamilegacy/kubectl:latest
+        imagePullPolicy: IfNotPresent
+        command:
+          - sh
+          - -c
+          - |
+            NS=$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)
```
timmarkhuff: Maybe another place we could refactor this into a shell script?

deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
```diff
@@ -46,6 +47,22 @@ spec:
           maxUnavailable: 0  # Aim for no downtime during rollout
 
       initContainers:
+      - name: wait-for-warmup
+        image: bitnamilegacy/kubectl:latest
```
timmarkhuff: This image is not maintained anymore. I wonder if it's a good dependency for us to have. 

I realized that we were already using `bitnamilegacy` prior to this PR, but maybe we shouldn't? I got Claude to summarize the history of the decision to use bitnami:

---

`bitnami/kubectl:latest` was originally introduced by Tom Faulhaber (`tomfaulhaber`) in [PR #184](https://github.com/groundlight/edge-endpoint/pull/184) ("Use Helm to deploy Edge Endpoint"), merged March 2025. Approved by `tyler-romero`, `robotrapta`, and `honeytung`.

The switch from `bitnami/` to `bitnamilegacy/` was done by `CoreyEWood` in [PR #288](https://github.com/groundlight/edge-endpoint/pull/288) ("Switch to bitnamilegacy kubectl image as a temporary fix for upcoming deprecation"), merged September 2025. Approved by `honeytung`, who noted it was LGTM as a quick fix.

Notably, Corey's PR #288 commit message explicitly calls this a **temporary fix** and says:

> This legacy catalog may be removed in the future, so we should switch to either self-hosting or using their secure image.

The current `bitnamilegacy` usage was always intended to be temporary, and `honeytung` (the author of PR #368) was the one who approved that temporary measure. The follow-up to migrate off it never happened.

---

My read is that we could just use the edge-endpoint image (which we already have) and cut out a 3rd party dependency. Maybe not critical for this PR, but something to consider.

deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
```diff
@@ -46,6 +47,22 @@ spec:
...
+        imagePullPolicy: IfNotPresent
+        command:
+          - sh
+          - -c
+          - |
+            NS=$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)
+            echo "Checking for warmup-inference-model job..."
```
timmarkhuff: `kubectl wait --for=condition=complete` will block for the full 30-minute timeout even if the job has already failed (reached its `backoffLimit`). It might be better to also poll for the `Failed` condition so it can bail out quickly. Something like:

```bash
while true; do
  if kubectl wait --for=condition=complete job/warmup-inference-model -n $NS --timeout=10s 2>/dev/null; then
    echo "Warmup complete."; break
  fi
  if kubectl wait --for=condition=failed job/warmup-inference-model -n $NS --timeout=1s 2>/dev/null; then
    echo "Warmup failed, proceeding anyway."; break
  fi
done
```

deploy/k3s/inference_deployment/inference_deployment_template.yaml
```diff
@@ -91,6 +108,8 @@ spec:
           value: /opt/models/pinamod
         - name: LOAD_ALL_PIPELINES  # Load only the pipelines that are needed for edge inference.
           value: "false"
+        - name: INFERENCE_FLAVOR
```
timmarkhuff, line 100: Not related to this PR, but doesn't hurt to add I supposed. We will deprecate this path soon anyway.

timmarkhuff (approved):
LGTM.

I left a few comments.

deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
honeytung: The `wait-for-warmup` now uses the edge-endpoint image. From discussion offline we found out that we cannot fully remove all `bitnami/kubectl:latest` images in favor of edge-endpoint image since we require a public available image to pull credentials from aws.

We have decided to keep `bitnami/kubectl:latest` for some parts (switched from `bitnamilegacy` as the support for `bitnami/kubectl:latest` seems to be back). We should revisit this when we work on reducing third party image sources for edge.

PR #369, opened 2026-04-01
Add httpEnabled flag to Helm chart to allow disabling HTTP port

honeytung:
## Summary

Adds a configurable `httpEnabled` Helm value (default: `true`) that controls whether the HTTP (non-TLS) port 30101 is exposed on the edge endpoint. This allows operators to enforce HTTPS-only access after enabling HTTPS on port 30143.

Also fixes the root URL redirect (`/` → `/status`) to use a relative path with `absolute_redirect off`, so it works correctly behind NodePort mappings where the external port differs from the internal NGINX port.

Resolves GL-1268.

## Usage

To disable HTTP and allow only HTTPS access:

```bash
helm install edge-endpoint deploy/helm/groundlight-edge-endpoint/ \
  --set httpEnabled=false \
  --set groundlightApiToken=<token>
```

Or to implicitly keep HTTP enabled (default behavior, no change needed):

```bash
helm install edge-endpoint deploy/helm/groundlight-edge-endpoint/ \
  --set groundlightApiToken=<token>
```

## Changes

When `httpEnabled` is set to `false`:

| Component | Behavior |
|-----------|----------|
| **Kubernetes Service** | HTTP NodePort (30101) is removed; only HTTPS NodePort (30143) remains |
| **NGINX config** | `listen 30101;` directive is removed; only `listen 443 ssl;` remains |
| **Container ports** | HTTP containerPort 30101 is removed from the nginx container |
| **Health probes** | Startup/readiness/liveness probes switch from HTTP:30101 to HTTPS:443 |
| **Helm tests** | Test pods use HTTPS with `--no-check-certificate` for self-signed certs |
| **NOTES.txt** | Connection instructions reflect HTTPS-only mode |

Additional fixes (apply regardless of `httpEnabled`):
- **NGINX redirect fix**: Root URL redirect (`/` → `/status`) now uses a relative path with `absolute_redirect off`, so it works correctly when the external NodePort (e.g., 30143) differs from the internal NGINX port (443)
- **Comment updates**: Clarified that `edgeEndpointPort` is only used when `httpEnabled` is true; fixed typo

When `httpEnabled` is `true` (default), behavior is identical to before this change — fully backward compatible.

## Files Changed

- `deploy/helm/groundlight-edge-endpoint/values.yaml` — Added `httpEnabled: true` with documentation; updated comments
- `deploy/helm/groundlight-edge-endpoint/templates/edge-deployment.yaml` — Conditional HTTP port in Service, container ports, and probe configuration
- `deploy/helm/groundlight-edge-endpoint/files/nginx.conf` — Conditional `listen 30101` directive; fixed root redirect to use relative path
- `deploy/helm/groundlight-edge-endpoint/templates/tests/test-local-connection.yaml` — HTTPS fallback for tests
- `deploy/helm/groundlight-edge-endpoint/templates/tests/test-cloud-connection.yaml` — HTTPS fallback for tests
- `deploy/helm/groundlight-edge-endpoint/templates/NOTES.txt` — Updated install notes for both modes

## Testing

Tested on G4 instance with `httpEnabled=false`:
- ✅ All pods (5/5 containers) spin up and pass health checks over HTTPS
- ✅ HTTP port 30101 is not accessible (connection refused)
- ✅ HTTPS port 30143 serves health checks (`/health/ready` returns `{"status":"ready"}`)
- ✅ Inference requests succeed over HTTPS (tested with `det_2txDBXmAtK9HoCQUilSUivHx0rd`)
- ✅ Status page loads correctly at `https://localhost:30143/status`
- ✅ Root redirect `https://localhost:30143/` → `https://localhost:30143/status` works correctly
- ✅ K8s Service only exposes HTTPS NodePort (30143), no HTTP port

[Slack Thread](https://axon.slack.com/archives/C09AH3HN6Q6/p1775060531124019?thread_ts=1775060531.124019&cid=C09AH3HN6Q6)


deploy/helm/groundlight-edge-endpoint/templates/tests/test-cloud-connection.yaml
```diff
@@ -13,10 +13,13 @@ spec:
     - name: wget
       image: busybox
       command: ['sh']
-      # Note that we use port 30101 here which is the internal port and not the port exposed to the host
```
timmarkhuff, line 16: Maybe update this comment instead of deleting it?

timmarkhuff:
This is a total nit, but in the description is says:

_Or to explicitly keep HTTP enabled (default behavior, no change needed):_
```
helm install edge-endpoint deploy/helm/groundlight-edge-endpoint/ \
  --set groundlightApiToken=<token>
```

This would be implicit, not explicit.

timmarkhuff:
Claude had the below suggestion, which based on my testing, seems like a good idea. I wasn't able to get the Python SDK to use the Edge Endpoint until I did this. 

The HTTPS instructions in NOTES.txt should mention the self-signed cert workaround, otherwise users will immediately hit SSL: CERTIFICATE_VERIFY_FAILED when they follow the instructions. Suggested addition after the HTTPS block:

NOTE: The edge endpoint uses a self-signed TLS certificate. Set DISABLE_TLS_VERIFY=1
or pass disable_tls_verification=True to the Groundlight() constructor.
This applies both when httpEnabled=false (HTTPS-only) and when httpEnabled=true (since HTTPS instructions are now always shown).

timmarkhuff (approved):
LGTM

PR #370, opened 2026-04-03
Network resilience: image preloading, pull policy, and S3 FUSE mount for model weights

honeytung:
## Summary

Improve network resilience for edge devices by eliminating runtime dependencies on DockerHub and replacing S3 sync with a FUSE-mounted S3 bucket for model weights.

### 1. Image pull policy & preloading (offline k3s support)
- Add `imagePullPolicy: IfNotPresent` to all DockerHub images in Helm templates (aws-cli, kubectl, alpine, busybox)
- Pin `alpine:latest` to `alpine:3.20` for consistency with preloaded images
- Add image preloading stages to Balena Dockerfiles (CPU, GPU, Jetson) using `crane` to pull images at build time (Note that the preloading steps *only* happens inside Balena build process, we do not do pre-loading when installing from helm directly)
- Copy preloaded tarballs to k3s airgap directory at runtime via `server.sh`

### 2. Replace S3 sync with mount-s3 FUSE mount
- **Remove** `warmup-inference-model` Job and `wait-for-warmup` init container
- **Remove** `sync-pinamod` init container from inference deployments
- **Add** `mount-s3` sidecar container to the edge-endpoint pod that FUSE-mounts `s3://pinamod-artifacts-public`
- Mount uses `--read-only` with local disk cache (`--cache`) for performance
- Mount on `/host-root/<path>` pattern to make FUSE mount visible to other pods via hostPath
- Inference pods read pinamod models and pretrained weights directly from the mount
- Set `HF_HUB_OFFLINE=1` and `TRANSFORMERS_OFFLINE=1` to prevent runtime downloads from HuggingFace
- Remove `aws-credentials` volume and `batch/jobs` RBAC from inference pods (no longer needed)
- Add `s3Mount` configuration values (bucket, region, mountPath, cachePath)

### 3. Files changed
- `Dockerfile` — install `mount-s3` + FUSE
- `deploy/bin/mount-s3.sh` — mount-s3 sidecar script (new)
- `deploy/bin/wait-for-warmup.sh` — deleted
- `deploy/helm/.../templates/edge-deployment.yaml` — add mount-s3 sidecar + host-root volume
- `deploy/helm/.../templates/warmup-inference-model.yaml` — deleted
- `deploy/helm/.../files/inference-deployment-template.yaml` — remove sync init containers, add env vars
- `deploy/helm/.../templates/service-account.yaml` — remove batch/jobs RBAC
- `deploy/helm/.../values.yaml` — add s3Mount config
- `deploy/balena-k3s/server/Dockerfile*` — add preloader stages
- `deploy/balena-k3s/server/server.sh` — copy preloaded images to k3s
- Various Helm templates — add `imagePullPolicy: IfNotPresent` to all DockerHub images except for edge-endpoint and inference images

## Test plan

Tested on G4 EC2 (Tesla T4, k3s v1.33.5):

- [x] mount-s3 sidecar starts and FUSE-mounts S3 bucket successfully
- [x] Host sees mounted files at `/opt/groundlight/edge/pinamod-mount/`
- [x] 9 detector inference pods start without warmup Job or sync-pinamod
- [x] Inference returns edge predictions (`from_edge: true`) for all 9 detectors
- [x] OD detector (`det_34Ws7...`) loads pinamod weights from FUSE mount
- [x] mount-s3 disk cache populates on first read (~83MB for OD weights)
- [x] No HuggingFace downloads (HF_HUB_OFFLINE=1 enforced)
- [x] `imagePullPolicy: IfNotPresent` set on all DockerHub images
- [x] Balena device test with preloaded images + restricted network


deploy/balena-k3s/server/Dockerfile
```diff
@@ -4,6 +4,44 @@
 # https://k3d.io/v5.7.4/usage/advanced/cuda/#building-a-customized-k3s-image
```
honeytung, line 4: The following Dockerfile changes are for preloading images with `crane`. This will also be used in GEP for balena deployments. Note that the three Dockerfile represents different architectures (`cpu`, `x86-GPU`, and `jetson`).

In GEP we kept those in jinja format and we generate the actual Dockerfile via Python. Here we kept full Dockerfile for each deployments so it is easier for people to reuse this for their custom deployment environments.

deploy/helm/groundlight-edge-endpoint/templates/k3s-memory-config.yaml
```diff
@@ -17,7 +17,8 @@ spec:
       hostPID: true
       containers:
       - name: k3s-memory-configurator
-        image: alpine:latest
+        image: alpine:3.20
```
honeytung, line 20: I modified `alpine` to use `3.20` instead of `latest` so that it matches what the `network-healer` is using below. I am not sure why they are different versions.

deploy/balena-k3s/server/Dockerfile
```diff
@@ -4,6 +4,44 @@
...
+    wget -q "https://github.com/k3s-io/k3s/releases/download/${K3S_VERSION}/k3s-airgap-images-${K3S_ARCH}.tar.zst" \
+        -O /preload/k3s-airgap-images.tar.zst || true
+
+# Download app utility images used by the Helm chart
+RUN ARCH=$(uname -m) && \
+    if [ "$ARCH" = "aarch64" ]; then PLAT="linux/arm64"; else PLAT="linux/amd64"; fi && \
+    crane pull --platform "$PLAT" docker.io/amazon/aws-cli:latest /preload/aws-cli-latest.tar && \
+    crane pull --platform "$PLAT" docker.io/bitnami/kubectl:latest /preload/kubectl-latest.tar && \
+    crane pull --platform "$PLAT" docker.io/alpine:3.20 /preload/alpine-3.20.tar && \
+    crane pull --platform "$PLAT" docker.io/busybox:1.35 /preload/busybox-1.35.tar
```
CoreyEWood: This pulls `busybox:1.35` but `edge-deployment.yaml` uses `busybox:1.36` for `apply-edge-config`. Should the `crane pull` command include `1.36` to cover this?

deploy/helm/groundlight-edge-endpoint/templates/edge-deployment.yaml
```diff
@@ -84,6 +84,28 @@ spec:
         command: ["/bin/sh", "/etc/groundlight/edge-config/apply-edge-config.sh"]
 
       containers:
+      - name: mount-s3
```
CoreyEWood, line 89: The other containers in this pod all have resources.requests.memory set to avoid eviction under memory pressure. Should mount-s3 have one as well?

deploy/balena-k3s/server/Dockerfile
```diff
@@ -4,6 +4,44 @@
 # https://k3d.io/v5.7.4/usage/advanced/cuda/#building-a-customized-k3s-image
 # https://github.com/k3s-io/k3s/blob/master/package/Dockerfile
 
+# -------------------------
+# Stage 1: Preload - download third-party images as tarballs for airgap/restricted network support
+# -------------------------
```
CoreyEWood, line 10: This preloader stage seems to be duplicated across Dockerfile, Dockerfile.gpu, and Dockerfile.jetson. Could we extract it into a shared Dockerfile or base image so the preload list only needs to be maintained in one place?

NETWORK-REQUIREMENTS.md
```diff
@@ -16,14 +16,10 @@ Depending on the deployment methods, there may be additional networking requirem
...
-- `*.us-west-2.amazonaws.com`: AWS access for inference image download
-- `*.s3.amazonaws.com`: AWS access for model binaries download
-- `*.sts.amazonaws.com`: AWS authentication
-- `*.huggingface.co`: Model download
-- `*.hf.co`: Hugging Face CDN
-- `*.docker.io`: Docker image download
-- `*.docker.com`: Docker image download
+- `*.groundlight.ai`: Groundlight endpoint for receiving detector metadata and temporary AWS credentials
+- `*.us-west-2.amazonaws.com`: AWS access for inference image download (ECR), model weights (S3), and credential refresh (STS)
+- `*.s3.amazonaws.com`: AWS S3 access for model weights via FUSE mount
```
CoreyEWood, line 21: Do you know why we need both of these instead of just the first one?

deploy/bin/wait-for-warmup.sh
CoreyEWood, line 1: Have we fully deprecated the non-helm deploy route? It looks like this script is still referenced in the old `deploy/k3s/inference_deployment_template.yaml`, so we might need to keep it around if anyone else is still using that deployment path.

deploy/bin/mount-s3.sh
```diff
@@ -0,0 +1,39 @@
...
+
+MOUNT_POINT="/host-root${MOUNT_PATH}"
+CACHE_DIR="/host-root${CACHE_PATH}"
+
+mkdir -p "$MOUNT_POINT" "$CACHE_DIR"
+
+# Copy AWS credentials to host for mount-s3 to use
+mkdir -p /host-root/root/.aws
+cp /root/.aws/credentials /host-root/root/.aws/credentials 2>/dev/null || true
+cp /root/.aws/config /host-root/root/.aws/config 2>/dev/null || true
```
CoreyEWood: Why is this needed if it already has access to the credentials here?

deploy/helm/groundlight-edge-endpoint/templates/edge-deployment.yaml
```diff
@@ -84,6 +84,28 @@ spec:
...
+          value: "{{ .Values.s3Mount.mountPath }}"
+        - name: CACHE_PATH
+          value: "{{ .Values.s3Mount.cachePath }}"
+        volumeMounts:
+        - name: host-root
+          mountPath: /host-root
+          mountPropagation: Bidirectional
+        - name: aws-credentials
+          mountPath: /root/.aws
+          readOnly: true
```
CoreyEWood, line 119: Should this container have a liveness probe? If the FUSE mount fails silently (e.g. a credentials issue), the pod would appear healthy but inference pods would see an empty directory. Claude says that something like checking `mountpoint -q /host-root/<mount_path>` could catch this.

deploy/helm/groundlight-edge-endpoint/templates/edge-deployment.yaml
```diff
@@ -84,6 +84,28 @@ spec:
         command: ["/bin/sh", "/etc/groundlight/edge-config/apply-edge-config.sh"]
 
       containers:
+      - name: mount-s3
+        image: *edgeEndpointImage
+        imagePullPolicy: "{{ include "groundlight-edge-endpoint.edgeEndpointPullPolicy" . }}"
+        securityContext:
+          privileged: true  # Required for FUSE mount
+        command: ["/bin/bash", "/groundlight-edge/deploy/bin/mount-s3.sh"]
```
CoreyEWood, line 97: Do you think we need to think about the ordering of this container vs the inference pods starting up at all? Most of the time it seems like it should be fine because we expect this to start before inference pods do, and it should run quickly. And maybe adding a liveness probe would cover the scenario where this doesn't succeed. But just want to flag in case there's other scenarios we should cover to avoid some kind of silent failure.

deploy/bin/mount-s3.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# Unmount if stale mount exists from a previous run
+if mountpoint -q "$MOUNT_POINT" 2>/dev/null; then
+    echo "Unmounting stale mount at $MOUNT_POINT"
+    umount "$MOUNT_POINT" || fusermount -u "$MOUNT_POINT" || true
+fi
+
+echo "Mounting s3://$S3_BUCKET at $MOUNT_POINT (cache: $CACHE_DIR, region: $S3_REGION)"
+mount-s3 "$S3_BUCKET" "$MOUNT_POINT" \
```
CoreyEWood, line 29: Not sure the current state of edge documentation but do you think we should add a short section about the mountpoint usage to a README somewhere?

CoreyEWood (commented):
Overall this looks great! Love how simple the S3 mounting is to set up. And I think ultimately the validation via testing is the most key part of it. I left comments which I think are worth looking at before merging but mostly about smaller portions of it.

deploy/balena-k3s/server/Dockerfile
honeytung: Good catch! Will add that to the list.

deploy/balena-k3s/server/Dockerfile
honeytung: We can, for GEP we used a jinja template. But for edge-endpoint since it is public facing I wanted to keep those architectures separate so it is easier for people to copy & modify it based on their needs.

With jinja templating we can make this cleaner, but it will require us to call a python script like GEP does to compile the actual `Dockerfile` at build time.

NETWORK-REQUIREMENTS.md
honeytung: I believe the AWS CLI may use global endpoints (`s3.amazonaws.com`, `sts.amazonaws.com`) without the region prefix even if specified, especially when talking to `sts` to get creds.

deploy/bin/wait-for-warmup.sh
honeytung: Good call, there are some people still using the non-helm deploy route. I will revert the script for now.

deploy/bin/mount-s3.sh
honeytung: I think I originally planned to run mount-s3 in host for testing, but I agree it should be necessarily, will test out and update it.

deploy/helm/groundlight-edge-endpoint/templates/edge-deployment.yaml
honeytung: Good call, will add it 🙂

deploy/helm/groundlight-edge-endpoint/templates/edge-deployment.yaml
honeytung: I think this is fine, usually `mount-s3` will start fairly quickly, but a liveness probe from the comment above should solve this issue. Also, the inference pod at startup will still need to pull the inference image so it will definitely be slower than the s3 mount.

deploy/bin/mount-s3.sh
honeytung: I think we can add that later, I think the current edge architecture docs only covers our escalation and ML logic but not the structure of the edge-endpoint.

deploy/balena-k3s/server/Dockerfile
```diff
@@ -4,6 +4,45 @@
...
+        -O /preload/k3s-airgap-images.tar.zst || true
+
+# Download app utility images used by the Helm chart
+RUN ARCH=$(uname -m) && \
+    if [ "$ARCH" = "aarch64" ]; then PLAT="linux/arm64"; else PLAT="linux/amd64"; fi && \
+    crane pull --platform "$PLAT" docker.io/amazon/aws-cli:latest /preload/aws-cli-latest.tar && \
+    crane pull --platform "$PLAT" docker.io/bitnami/kubectl:latest /preload/kubectl-latest.tar && \
+    crane pull --platform "$PLAT" docker.io/alpine:3.20 /preload/alpine-3.20.tar && \
+    crane pull --platform "$PLAT" docker.io/busybox:1.35 /preload/busybox-1.35.tar && \
+    crane pull --platform "$PLAT" docker.io/busybox:1.36 /preload/busybox-1.36.tar
```
CoreyEWood: It looks like we use 1.35 and 1.36 in one place each, I wonder if we can switch to using just one version to avoid having to pull both images? Unless there's a reason we need both 1.35 and 1.36 for those specific spots.

deploy/balena-k3s/server/Dockerfile
CoreyEWood: Okay gotcha, makes sense. In this case we'll just have to remember to keep all of these in sync.

CoreyEWood: Maybe add a note in this section that it should be kept in sync across all of these when making changes, so that people in the future are aware.

deploy/bin/wait-for-warmup.sh
CoreyEWood: Sounds good. And from looking at it, it seems to me like nothing else here should break non-helm deployments hopefully. Do you think we should test that explicitly to be safe? Or are you confident about that?

CoreyEWood (approved):
Changes look good! Left a few more comments but then should be ready to merge.

deploy/bin/wait-for-warmup.sh
honeytung: Yea I think this is fine, since we did not touch any `k3s` templates. We can communicate with the people still using the old `k3s` method to switch over soon so we can fully deprecate it.

PR #372, opened 2026-04-07
Add profiling system to edge-endpoint

f-wright:
This PR adds a profiling system to edge-endpoint which will give us better visibility into which components of inference are taking the most time, for future optimization. A follow up PR will extend the profiling system into the inference pod.

## Overview of changes

  - Adds an opt-in request-level tracing system that records per-stage timing for every inference request as JSONL to disk
  - Uses middleware + contextvar + `@trace_span` decorators — no changes to `image_queries.py`, adding new trace points is a one-line decorator
  - Sets up tracing in the inference pod by propagating trace headers (`X-GL-Trace-Id`, `X-GL-Parent-Span-Id`) to inference pods for cross-container distributed tracing and adding`device-info-volume` mount to inference pod template so inference pods can write traces to the shared volume

  ## How it works

  When `enableProfiling` is set to `true` in the Helm deployment, middleware creates a trace per request and sets a contextvar. Functions decorated with `@trace_span` automatically create child spans when a tracer is active. Traces are written as JSONL to `/opt/groundlight/device/edge-profiling/traces/` with 5-minute file rotation and 24-hour automatic cleanup.

  ## Traced spans
```
  request                              ← middleware (full request lifecycle)
  ├── get_detector_metadata            ← cache hit vs cloud API round-trip
  ├── inference_is_available           ← health check cache hit vs cold check
  ├── primary_inference                ← HTTP to primary inference pod
  ├── oodd_inference                   ← HTTP to OODD inference pod (parallel)
  ├── get_inference_result             ← result parsing + OODD confidence adjustment
  └── escalate_to_cloud               ← cloud escalation (when triggered)
```
Tracing will be added within the inference pods in a follow up PR for the inference server. 

  ## Deploy

  ```bash
  helm upgrade -i edge-endpoint ... --set enableProfiling=true
```

  ## Test plan

  - New unit tests covering the profiling module
  - All existing edge tests pass
  - Ran end to end on a g4 and observed successful profiling

### CODE REVIEW

File: deploy/helm/groundlight-edge-endpoint/values.yaml [hunk 1]
```diff
@@ -24,6 +24,11 @@ inferenceTag: ""
 # an object detection or counting model. 
 useMinimalImage: false
 
+# Enable request-level tracing profiling for the edge endpoint.
+# When enabled, per-request trace data (span timings, annotations) is written to
+# /opt/groundlight/device/edge-profiling/traces/ as JSONL files. Disabled by default.
```
timmarkhuff: I don't think the traces directory exists anymore

app/core/edge_inference.py
timmarkhuff: From Claude:

"Non-parallel inference path has no _submit_primary_inference span

This calls submit_image_for_inference directly instead of the traced _submit_primary_inference wrapper. So there's no _submit_primary_inference span in the trace tree for the non-parallel path. The trace headers are still propagated (good), but the timing of the primary inference HTTP call won't show up as a named span. May be intentional, but traces from minimal-image deployments will be less informative."

This seems like it might be worth considering as we move towards minimal inference.

timmarkhuff:
The device-info-volume mount was added to the Helm inference template but not to deploy/k3s/inference_deployment/inference_deployment_template.yaml, so inference-pod tracing won't work on Balena deployments. Low priority since this PR doesn't add profiling for inference containers, but maybe something to consider in a subsequent PR.

timmarkhuff (approved):
LGTM. I left a few comments. 

It would be nice to have a way to visualize the profiling results. That could be another PR.

I noticed code in the inference path was touched. Have you run the benchmarking script to ensure throughput was not adversely affected? Might be a good idea.

f-wright:
> It would be nice to have a way to visualize the profiling results. That could be another PR.

Agreed! That work is already planned for a follow up PR, coming soon.

> I noticed code in the inference path was touched. Have you run the benchmarking script to ensure throughput was not adversely affected? Might be a good idea.

Thanks for the suggestion, I ran the benchmarking script for a binary detector with the release inference image, and with these changes with profiling turned off and on. All reached the same throughput. I also repeated the same process for a count detector and got the same results.

PR #373, opened 2026-04-10
Benchmarking VRAM and RAM

timmarkhuff:
1. Adds `measure_ram_and_vram_usage.py`, a benchmarking script that measures per-detector VRAM and RAM on a running Edge Endpoint across a YAML-defined matrix of `(mode, pipeline, n, image_size)` combinations. Supports `--resume <run-dir>` for long sweeps.
2. Adds `plot_ram_and_vram_usage.py`, which renders a side-by-side VRAM/RAM horizontal-bar figure grouped by mode and pipeline.
3. Adds unit tests for load-testing helpers and runs them in CI.
4. Adds `provision_detectors`, a faster way to provision many detectors by priming all upfront and waiting for cloud training in parallel.
5. Updates `groundlight_helpers`: priming label count now scales with `n` to ensure that detectors with high n values get sufficient labels for each class/number of bounding boxes. Also, `configure_edge_endpoint` sets a faster refresh rate, so that Edge Endpoints are faster to respond to changes in configuration.

The script outputs a plot like this:
image


load-testing/README.md
```diff
@@ -100,3 +100,23 @@ Below are some commands that are commonly run to evaluate the performance of the
...
+
+#### Purpose
+Measures per-detector VRAM and RAM on a running Edge Endpoint across a YAML-defined matrix of `(mode, pipeline, n, image_size)`.
+
+#### Usage
+```
+uv run python measure_ram_and_vram_usage.py PIPELINES_YAML --device-name DEVICE [options]
+```
+
+See `benchmark_pipelines.example.yaml` for the input format. Each mode block requires `image_sizes` and `pipelines`; non-BINARY modes may also set `n`. The script expands the cartesian product into one measurement per combination.
```
honeytung: I think you mean `benchmark_pipelines.yaml` since that is the file added in this PR?

honeytung (approved):
LGTM! Left a tiny comment.

load-testing/README.md
timmarkhuff: Good catch. Fixed.

PR #374, opened 2026-04-13
Fix 503s during inference pod rollouts

timmarkhuff:
## Summary

- Fix intermittent 500/503 errors during inference pod rolling updates by adding a `preStop` lifecycle hook to the inference container.
- Move the Deployment `strategy` block from `spec.template.spec` (Pod spec, where it was silently ignored) to `spec` (Deployment spec, where it actually takes effect).
- Add `rollouts_under_inference_load.py` load test script to reproduce and validate the fix.

## Problem

During inference pod rollouts, clients intermittently received `500 Internal Server Error` with `Connection refused`. This happened because Kubernetes pod termination and endpoint removal are concurrent, not sequential:

> "At the same time as the kubelet is starting graceful shutdown of the Pod, the control plane evaluates whether to remove that shutting-down Pod from EndpointSlice objects"
>
> -- [Pod Lifecycle: Termination of Pods](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination)

When the old pod received SIGTERM, uvicorn immediately closed its listening socket. But kube-proxy had not yet updated its iptables rules, so requests were still routed to the now-closed socket.

## Solution

Added a `preStop` hook (`sleep 10`) to the inference container. This delays SIGTERM, giving kube-proxy time to remove the pod from iptables before the process stops accepting connections:

> "If one of the Pod's containers has defined a `preStop` hook [...] the kubelet runs that hook inside of the container."
>
> "The kubelet triggers the container runtime to send a TERM signal to process 1 inside each container."
>
> -- [Pod Lifecycle: Termination of Pods](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination)

### Strategy block placement fix

The `strategy` block was previously placed under `spec.template.spec` (the [PodSpec](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#PodSpec)), where it is not a recognized field and was silently ignored. Moved it to `spec` (the [DeploymentSpec](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/deployment-v1/#DeploymentSpec)), where `strategy` is defined. This ensures the explicit `maxUnavailable: 0` takes effect, which matters if replicas are ever increased beyond 1 (the default `maxUnavailable: 25%` would allow unavailable pods at higher replica counts).

## Test plan

- [x] Run `load-testing/rollouts_under_inference_load.py` which submits inference queries as fast as possible while triggering model rollouts via label submission. Passes if 3 minutes of continuous inference complete with zero errors.
- [x] Validated locally: 8,317 inferences over 180s through multiple rollouts with zero errors.


deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
```diff
@@ -40,11 +45,6 @@ spec:
       serviceAccountName: edge-endpoint-service-account
       imagePullSecrets:
       - name: registry-credentials
-      strategy:
```
timmarkhuff, line 43: This has always been in the wrong place (and therefore ignored), but because the defaults are so similar to this, we haven't noticed.

deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
```diff
@@ -23,6 +23,11 @@ metadata:
     name: placeholder-inference-deployment-name
 spec:
   replicas: 1
+  strategy:
```
timmarkhuff, line 26: This is now in the correct place, although it doesn't seem to materially change the behavior of the system, and doesn't fix the 500s we see during rollouts

load-testing/rollouts_under_inference_load.py
```diff
@@ -0,0 +1,164 @@
...
+DETECTOR_NAME = 'Rollout Under Load Test'
+DETECTOR_GROUP_NAME = 'Edge Endpoint Load Testing'
+LABEL_INTERVAL_SEC = 30.0
+LABELS_PER_BATCH = 3
+MIN_STARTING_LABELS = 30
+CONFIDENCE_THRESHOLD = 0.0
+VICTORY_DURATION_SEC = 3 * 60
+
+
+def disable_sdk_retries(gl: ExperimentalApi) -> None:
```
timmarkhuff, line 22: removing all retries so that errors are never hidden

load-testing/pyproject.toml
```diff
@@ -2,10 +2,10 @@
 name = "load-testing"
 version = "0.1.0"
 description = "Scripts for testing the performance of the Groundlight Edge Endpoint"
-requires-python = ">=3.13"
+requires-python = ">=3.13,<3.14"
```
honeytung: Any reason that we are limiting python version higher than `3.14`?

deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
honeytung: Nice catch, interesting that k3s rollout did not just yell at you and error out.

honeytung (approved):
LGTM!

deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
timmarkhuff: It seems that since the inference pods are created via the k3s python SDK, the errors/warnings didn't surface

load-testing/pyproject.toml
timmarkhuff: No good reason. I just reverted that.

PR #379, opened 2026-04-14
Various updates to benchmarking scripts

timmarkhuff:
## Summary 
This PR makes some usability improvements, mainly targeted at `multiple_client_throughput_test.py`, but other scripts were also refactored in the same fashion so that they stay up-to-date. The most useful test to us at this point is `multiple_client_throughput_test.py`, so I focused my manual testing efforts on that. 

## Changes
1. makes use of new python-sdk features to automatically configure the Edge Endpoint with the correct detector with the correct Edge Config (gl.edge.set_config). 
2. removes some old cruft that hasn't been used in a while (dog images, unused config files, etc.)
3. changes the logic for polling for a detector to complete training. The old method was flawed because it polled the `evaluation` endpoint, which could only give us information on the active pipeline (which likely isn't the edge pipeline). Now that we have a `pipelines` endpoint we can get the information we really need, which is the training details for the Edge pipeline. 
4. Adds support for BOUNDING_BOX detectors and centralizes logic for provisioning detectors and generating images for detectors. 
5. `single_client_fps_test.py` was removed because `multiple_client_throughput_test.py` tests all the same things (and more).
6. `generate_repeated_rollouts.py` was also removed. I spent a little time trying to modernize it and get it working again, but it didn't feel worthwhile. It tests a scenario that isn't recommended anymore (using a bunch of unconfigured detectors), so it doesn't feel like it's worth keeping around anymore.


load-testing/groundlight_helpers.py
```diff
@@ -15,15 +14,14 @@
 from urllib.parse import urlparse
 
 CLOUD_ENDPOINT_PROD = 'https://api.groundlight.ai/device-api'
+SUPPORTED_DETECTOR_MODES = {"BINARY", "COUNT", "BOUNDING_BOX"}
```
f-wright, line 17: Is multiclass not supported? That seems like something we'll want soon. No need to add it to this PR if you think it'd be better to add later though.

load-testing/groundlight_helpers.py
```diff
@@ -274,96 +273,123 @@ def assert_configured_edge_pipeline_matches_provided(
...
+            max_count=10,
+            group_name=group_name,
+            edge_pipeline_config=edge_pipeline_config,
+        )
+    elif detector_mode == "BOUNDING_BOX":
+        detector = get_or_create_bounding_box_detector(
+            gl,
+            name=detector_name,
+            class_name="circle",
+            max_num_bboxes=10,
```
f-wright, line 335: Do we want to make this configurable? I imagine some of the lenses will use a higher max count, and for some pipelines I think the inference time will be higher when there's more objects. Same on count mode.

load-testing/image_helpers.py
```diff
@@ -101,6 +106,18 @@ def generate_random_image(
             class_name=class_name,
             max_count=max_count,
         )
+    elif detector_mode == 'BOUNDING_BOX':
+        config = detector.mode_configuration
+        class_name = config["class_name"]
+        max_num_bboxes = int(config.get("max_num_bboxes", 10))
+        image, _, rois = generate_random_count_image(
```
f-wright: Maybe rename `generate_random_count_image` to something more general for objects?

load-testing/single_client_fps_test.py
```diff
@@ -1,200 +0,0 @@
-from groundlight import ExperimentalApi
```
f-wright, line 1: This file is in the load testing README, maybe along with other removed files. Can you update the README before merging this?

load-testing/groundlight_helpers.py
```diff
@@ -274,96 +273,123 @@ def assert_configured_edge_pipeline_matches_provided(
...
-                f"  Loaded edge:     {loaded!r}"
-            )
-        loaded = (get_detector_edge_metrics(gl, detector_id) or {}).get("pipeline_config")
-        if _pipeline_configs_equal(expected, loaded):
-            return
-        time.sleep(5)
+    return detector
 
 
-def wait_for_ready_inference_pod(
```
f-wright, line 314: What is this functionality replaced by? The waiting for training doesn't seem to require a ready inference pod, or am I missing something?

load-testing/groundlight_helpers.py
```diff
@@ -274,96 +273,123 @@ def assert_configured_edge_pipeline_matches_provided(
...
 ) -> None:
-    """Raises if the pipeline loaded on the edge does not match the expected config."""
-    loaded = (get_detector_edge_metrics(gl, detector_id) or {}).get("pipeline_config")
-    if not _pipeline_configs_equal(expected_pipeline_config, loaded):
-        raise RuntimeError(
-            f"The pipeline config provided does not match the pipeline loaded for detector {detector_id}. "
-            "This can happen if the detector's pipeline config was changed after creation (e.g. via admin).\n"
-            f"  Provided: {expected_pipeline_config!r}\n"
-            f"  Loaded:   {loaded!r}"
+    """Push an edge config for one or more detectors and wait for readiness."""
```
f-wright, line 279: What waits for readiness here? That functionality seems to be missing to me

load-testing/rollouts_under_inference_load.py
```diff
@@ -56,58 +56,24 @@ def add_label(gl_cloud: Groundlight, detector: Detector) -> None:
...
-    # Configure edge with NO_CLOUD
-    edge_config = EdgeEndpointConfig()
-    edge_config.add_detector(detector, NO_CLOUD)
-    print('Pushing edge config (NO_CLOUD)...')
-    gl.edge.set_config(edge_config)
-    print('Edge config applied, inference pod ready.')
+
+    glh.configure_edge_endpoint(gl, detector)
+
+    # There is a bug where sometimes inference fails if the inference pod came online very recently.
```
f-wright, line 74: This seems like what the readiness poll should fix? I think we should poll for actual readiness, or submit a couple of queries until they work properly rather than assuming 3 seconds will be enough

f-wright (commented):
A couple of small comments, and I'm concerned about taking out the readiness check. Maybe that functionality was added in somewhere else that I missed, but if not, I think we should bring it back before merging this

load-testing/groundlight_helpers.py
timmarkhuff: Waiting for readiness is implicit in gl.edge.set_config

load-testing/rollouts_under_inference_load.py
timmarkhuff: Yeah, the readiness poll _mostly_ fixes this issue, but I was observing a behavior where sometimes the readiness polling finishes, and then for some reason the detector suddenly isn't ready anymore. 

I didn't dig into it too deeply, but I did note that the period of unreadiness is short, usually sub-second. I suspect something weird is going on with the readiness cache and the multiple Edge Endpoint workers (which each have their own cache). 

I want to dig in more, but in a separate PR.

load-testing/groundlight_helpers.py
timmarkhuff: It is replaced by:
edge_pipeline_is_sufficiently_trained, which waits until the edge pipeline finishes training in the cloud
gl.edge.set_config, which waits for the detector to be loaded on edge, and ready

load-testing/groundlight_helpers.py
timmarkhuff: It should be. It slipped my mind. I'll add in a follow up PR.

load-testing/groundlight_helpers.py
timmarkhuff: This is a good idea. I'll add it in a follow on PR.

load-testing/image_helpers.py
timmarkhuff: Good call. I will rename it to make it appropriately generic.

load-testing/single_client_fps_test.py
timmarkhuff: Good catch. Fixed.

load-testing/groundlight_helpers.py
f-wright: Ah I see, didn't realize the waiting was handled in the SDK. Sounds good! The naming made me think it didn't include waiting, maybe add a comment or update the SDK name in a separate PR?

load-testing/rollouts_under_inference_load.py
f-wright: Huh ok, that's strange. Sure, we can put that into a separate PR

f-wright (approved):

PR #380, opened 2026-04-16
Add Marimo dashboard for profiling trace visualization

f-wright:
Adds an interactive notebook for exploring JSONL trace data produced by the profiling system. Designed to be run on-demand and easily expandable. 

The notebook contains summary stats, per-span latency table, box plots + histograms with p50/p95/p99 markers, per-span scatter over time, request-duration scatter per detector, throughput chart, and a waterfall chart for investigating a single trace.


app/profiling/dashboard.py
```diff
@@ -0,0 +1,662 @@
+import marimo
```
f-wright, line 1: I decided to make this a marimo notebook since we'll be using the profiling traces primarily for on demand investigation, and it seemed like a good fit for making it interactive and easily expanded.

In the future if we wanted to, we could move this out of a notebook into a something more like the status page, but I think this works for now since profiling will only be enabled at very specific points.

app/api/routes/image_queries.py
```diff
@@ -33,11 +34,13 @@ async def validate_content_type(request: Request) -> str:
     return request.headers.get("Content-Type", "")
 
 
+@trace_span
```
f-wright, line 37: As I was reviewing the profiling results, I decided to trace some more functions to try to get a more full picture of the request. There might still be more to add later on.

app/profiling/README.md
```diff
@@ -0,0 +1,135 @@
...
+```bash
+ENABLE_PROFILING=true
+```
+
+When enabled, the profiling middleware creates a trace per request. Functions decorated with `@trace_span` automatically create child spans. Traces are written as JSONL to `/opt/groundlight/device/edge-profiling/` with 5-minute file rotation and 24-hour automatic cleanup.
+
+## Traced Spans
+
+```
+request                              <- middleware (full request lifecycle)
```
timmarkhuff: Having this in the readme might be hard to maintain (two places to change when you want to adjust profiling).

app/profiling/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Prerequisites
+
+Marimo and plotly live in an **optional** `profiling` dependency group, so they are **not** installed by default. To install them:
+
+```bash
+poetry install --with profiling --no-root
+```
+
+This keeps the ~50MB of dashboard dependencies out of everyone else's environment while letting anyone who wants the dashboard opt in.
```
timmarkhuff: Nit, but this seems like a comment that belongs in a PR description or a code comment somewhere, not in a readme. Readme should be more focused on how to use profiling

app/profiling/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+> **Note**: The production Docker image is built with `--without dev --without lint` (and does not include the `profiling` group either), so marimo is **not** installed in the `edge-endpoint` container. To view traces from a device, copy the trace files to your workstation (see [Custom Trace Data Directory](#custom-trace-data-directory)) or run `poetry install --with profiling --no-root` inside the container before launching the dashboard.
+
+### Launching the Dashboard
```
timmarkhuff: When I first read this, I didn't understand there were two optional paths here (run locally or on edge device). I think that could be more explicit with language like:

### Launching the Dashboard
The dashboard reads JSONL trace files from a directory on disk. There are two
common ways to run it:
#### Option 1: On your laptop (recommended for ad-hoc investigation)
Copy the trace files off the device first...
[PROFILING_TRACES_DIR example]
#### Option 2: On the edge device itself
Useful when you want to watch traces live without copying files...
[--host 0.0.0.0 example]

app/profiling/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+```bash
+poetry run marimo run app/profiling/dashboard.py
+```
+
+Marimo starts a web server on port 2718 (or the next open port if that's in use) and prints the URL. The dashboard is **read-only** in `run` mode — to edit cells, use `marimo edit` (see [Editing the Notebook](#editing-the-notebook) below). Stop the server with `Ctrl+C`.
+
+### On an Edge Device
+
+After installing dev dependencies (see Prerequisites), SSH into the device or exec into the `edge-endpoint` container, then:
```
timmarkhuff: Did you try these steps on a headless/edge device? I ran into an issue on my G4 where I got `DBusErrorResponse / PromptDismissedException / Failed to create the collection errors` while doing poetry install. I worked around it with: 
```bash
PYTHON_KEYRING_BACKEND=keyring.backends.fail.Keyring poetry install --with profiling --no-root
```

Claude suggests that this error would likely happen on any edge device. I'm not sure if that's true, but maybe worth adding something about this in the readme, in case someone runs into it. 

The local/laptop approach seems confusing to me, because the user has to find a way to copy the trace file to their laptop and point the dashboard at it; it's a lot of hoops to jump through. Maybe the edge device workflow should be preferred?

pyproject.toml
```diff
@@ -21,6 +21,7 @@ pyyaml = "^6.0.2"
 SQLAlchemy = "2.0.35"
 svix-ksuid = "^0.6.2"
 uvicorn = {version = "^0.31.0", extras = ["standard"]}
+[tool.poetry.group.dev]
```
timmarkhuff, line 25: (from Claude) `pyproject.toml` line 24: The empty `[tool.poetry.group.dev]` header was removed in 2f85210 but the auto-format bot in 7874cf8 re-added it. `poetry check` passes so it's not a hard bug, just cosmetic — but the commit message claims it's gone when it isn't, and the autoformatter will re-add it on every push. Either accept it or figure out what in the formatter pipeline (toml-sort?) is normalizing this.

app/profiling/context.py
```diff
@@ -18,10 +19,30 @@ def get_current_span():
...
     """Decorator that creates a child span (named after the function) when a tracer is active.
 
-    When no tracer is set in the context (profiling disabled or outside a traced request),
-    the decorated function is called directly with ~50-100ns overhead (one ContextVar.get()).
+    Works on both sync and async functions. When no tracer is set in the context
+    (profiling disabled or outside a traced request), the decorated function is called
+    directly with ~50-100ns overhead (one ContextVar.get()).
     """
 
+    if asyncio.iscoroutinefunction(func):
```
timmarkhuff, line 27: (from Claude) `app/profiling/context.py` line 27: The new async branch has zero direct unit tests — `test/profiling/test_context.py` only covers the sync wrapper. The async path is exercised indirectly through the request handler decorations, but it's the riskiest 13 lines added by the PR. Worth mirroring the existing sync tests (noop / creates span / finished on exception / nested) for the async path.

timmarkhuff (approved):
Looks good! I left a few comments about the readme. 

I had one other miscellaneous observation: 
1. In the "Select Trace" dropdown of the Trace Waterfall, I see some options that say unknown (e.g. `2026-04-22T21:00:07 | unknown | 0.3ms | 4c057cba342a`). Usually, they look like `2026-04-22T21:00:07 | det_3CjFaAlZMmW3ukQ10YswlhjJ4Xl | 350.0ms | f7f682ee6c30`. So I guess the unknowns are traces
on some function that is not related to a detector? I'm not sure what to make of these traces, as they seem to alwayst just include `request`. Should we filter them out of this dropdown? When might they be useful?

app/profiling/context.py
f-wright: Added these!

pyproject.toml
f-wright: I think it's fine for this to stay. I removed it earlier because I thought claude added it, not the auto-format bot

app/profiling/README.md
f-wright: I made the edge device workflow preferred, and added the issue you saw to the troubleshooting section. I was using a slightly different approach (guided by claude) to launch it, which I made the preferred method

f-wright:
> 1. In the "Select Trace" dropdown of the Trace Waterfall, I see some options that say unknown (e.g. `2026-04-22T21:00:07 | unknown | 0.3ms | 4c057cba342a`). Usually, they look like `2026-04-22T21:00:07 | det_3CjFaAlZMmW3ukQ10YswlhjJ4Xl | 350.0ms | f7f682ee6c30`. So I guess the unknowns are traces
>    on some function that is not related to a detector? I'm not sure what to make of these traces, as they seem to alwayst just include `request`. Should we filter them out of this dropdown? When might they be useful?

Yeah, those traces are not on the main inference path. I think I'd rather keep them in the dropdown in case we see some of them are at the top for longest, but the option to exclude them is why I added the filtering option for including a particular trace value. If you select primary inference in that filter, the unknown traces will be filtered out.

PR #382, opened 2026-04-21
Add support for MULTI_CLASS detectors in load testing scripts

timmarkhuff:
## Additions
1. Support for MULTI_CLASS detectors was added to `groundlight_helpers.py` and applied to `multiple_client_throughput_test.py`
2. A single integer knob `n` was introduced to control the mode-specific detector parameter: `max_count` (COUNT), `max_num_bboxes` (BOUNDING_BOX), or `num_classes` (MULTI_CLASS). Users can now tweak this when provisioning a detector, benchmarking, etc. Usage looks like this:
    ```bash
    uv run python multiple_client_throughput_test.py MULTI_CLASS 6
    ```


load-testing/README.md
```diff
@@ -65,21 +59,6 @@ Tests the Edge Endpoint's ability to handle concurrent client load by spawning m
 uv run python multiple_client_throughput_test.py DETECTOR_MODE [options]
 ```
 
-#### Options
```
timmarkhuff, line 68: This is duplication of code in the scripts. Users should run the script with --help to see the definitive list of what options are available.

load-testing/groundlight_helpers.py
```diff
@@ -288,6 +310,37 @@ def configure_edge_endpoint(
...
+        # mode_configuration.
+        return get_detector_mode_default_cardinality(mode)
+
+    config = detector.mode_configuration
+    if mode == "COUNT":
+        return int(config["max_count"])
+    if mode == "BOUNDING_BOX":
+        return int(config["max_num_bboxes"])
+    if mode == "MULTI_CLASS":
+        return len(config["class_names"])
```
f-wright: Does the config not have `num_classes` directly? I thought it did, but I could be wrong

load-testing/groundlight_helpers.py
```diff
@@ -288,6 +310,37 @@ def configure_edge_endpoint(
...
+    """Return the cardinality of the provided detector."""
+    mode = detector.mode
+    if mode == "BINARY":
+        # BINARY's cardinality is structurally fixed at 2, so there's nothing to read from
+        # mode_configuration.
+        return get_detector_mode_default_cardinality(mode)
+
+    config = detector.mode_configuration
+    if mode == "COUNT":
+        return int(config["max_count"])
```
f-wright: I think `max_count` is a little different from cardinality, since the resulting num classes is actually `max_count` + 2 (including greater than max and no objects)

load-testing/groundlight_helpers.py
```diff
@@ -288,6 +310,37 @@ def configure_edge_endpoint(
...
+    if mode == "BOUNDING_BOX":
+        return int(config["max_num_bboxes"])
```
f-wright: `max_num_bboxes` isn't really the cardinality of a bbox detector either,  there are 3 (4 if you count unclear) different labels possible, not the number of bboxes

load-testing/groundlight_helpers.py
```diff
@@ -323,7 +394,7 @@ def provision_detector(
             gl,
             name=detector_name,
             class_name="circle",
-            max_count=10,
+            max_count=resolved_cardinality,
```
f-wright: I guess the cardinality is still passed properly into the detector creation though, just a bit of a confusing function name. Do we even need to pass these values in, or are they the default through the SDK? 

I kind of think it makes more sense to keep the numbers here for count and bounding box mode, and then have a function just for getting the right number of classes for multiclass mode. What do you think?

load-testing/multiple_client_throughput_test.py
```diff
@@ -278,5 +283,14 @@ def main(  # noqa: PLR0913
     parser.add_argument("--image-width", type=int, default=640)
     parser.add_argument("--image-height", type=int, default=480)
     parser.add_argument("--edge-pipeline-config", type=str, default=None, help="Edge pipeline configuration name.")
+    parser.add_argument(
+        "--cardinality", type=int, default=None,
+        help=(
+            "Size of the detector's output/label space. Maps to max_count for COUNT, "
```
f-wright: This isn't correct for count and bbox mode

f-wright (commented):
Thanks for adding multiclass support! I left some questions about the approach to cardinality, I think we'll want changes there before we merge this.

load-testing/groundlight_helpers.py
timmarkhuff: `num_classes` is there. We should use it. I fixed it.

load-testing/groundlight_helpers.py
timmarkhuff: I tried having them be separate args, and it seemed a little messy. 

I have made another attempt to unify them under a single arg `n`, which gets explained in the relevant docstring and CLI helper text. Let me know what you think.

load-testing/multiple_client_throughput_test.py
timmarkhuff: Good call. I will move away from the `cardinality` approach.

load-testing/groundlight_helpers.py
```diff
@@ -288,9 +310,44 @@ def configure_edge_endpoint(
...
+    if n is None:
+        return
+    if n < 2:
+        raise ValueError(f"n must be >= 2 (got {n}).")
+    if detector_mode == "BINARY" and n != 2:
+        raise ValueError("n must be 2 for BINARY detectors (or omitted).")
+
+
 def provision_detector(
-    gl: ExperimentalApi,
```
timmarkhuff, line 292: I realized we don't actually need two clients anymore. This was left over from a previous implementation.

load-testing/groundlight_helpers.py
```diff
@@ -288,9 +310,44 @@ def configure_edge_endpoint(
...
+def default_n_for_mode(detector_mode: str) -> int:
+    """Return the default `n` value to use when none is provided for this mode."""
+    if detector_mode == "BINARY":
+        return 2
+    if detector_mode == "COUNT":
+        return 10
+    if detector_mode == "BOUNDING_BOX":
+        return 10
+    if detector_mode == "MULTI_CLASS":
+        return 4
```
f-wright, line 335: Where's this value coming from?

load-testing/groundlight_helpers.py
```diff
@@ -288,9 +310,44 @@ def configure_edge_endpoint(
...
+    raise ValueError(f"Unsupported detector mode: {detector_mode}")
+
+
+def validate_n_for_mode(detector_mode: str, n: int | None) -> None:
+    """Raise ValueError if `n` is incompatible with this detector mode. No-op when n is None."""
+    if n is None:
+        return
+    if n < 2:
```
f-wright: `max_count` and `max_num_bboxes` can both be 1

load-testing/multiple_client_throughput_test.py
```diff
@@ -253,11 +259,12 @@ def main(  # noqa: PLR0913
         steady_rps=throughput_summary.maximum_steady_rps,
     )
 
-    metadata = _collect_run_metadata(gl, detector, detector_mode, image_width, image_height)
```
f-wright, line 256: Was this intentional? I imagine we can see it in the detector, but is it equally accessible?

load-testing/multiple_client_throughput_test.py
```diff
@@ -269,8 +276,25 @@ def main(  # noqa: PLR0913
...
+        help=f"Detector mode to test. One of: {', '.join(sorted(glh.SUPPORTED_DETECTOR_MODES))}.",
+    )
+    parser.add_argument(
+        "n", type=int, nargs="?", default=None,
+        help=(
+            "Mode-specific integer knob.\n"
+            "  COUNT:        sets max_count\n"
+            "  BOUNDING_BOX: sets max_num_bboxes\n"
+            "  MULTI_CLASS:  sets num_classes\n"
+            "  BINARY:       number of classes, must be 2\n"
```
f-wright, line 294: Maybe we should just say should be omitted for binary

f-wright (approved):
Looks good! Thanks Tim

load-testing/multiple_client_throughput_test.py
timmarkhuff: It was intentional. I realized that you could just extract the mode from the detector, and that seemed cleaner.

load-testing/multiple_client_throughput_test.py
timmarkhuff: Maybe, I thought about that, but then I felt bad about erroring if the user enters 2. It's technically correct, so I thought we should accept it.

load-testing/groundlight_helpers.py
timmarkhuff: These are arbitrary defaults that I selected.

PR #383, opened 2026-04-22
Stamp primary+OODD MLB KSUIDs into edge_result metadata

honeytung:
## Summary

- `run_inference()` in `app/core/edge_inference.py` now attaches the currently-loaded primary (and OODD, when applicable) MLB KSUID to `output_dict` so cloud persists them on `Posicheck.metadata.edge_result`. Removes the need to infer edge MLBs by cross-referencing cloud supersession in OpenSearch for edge-vs-cloud debugging.
- Stamping is wrapped in `try/except` and guards each field on non-None reads, so a missing `model_id.txt` (first-boot race, disk issue, etc.) just omits the key without affecting inference.
- Helm: adds a second `/opt/groundlight/edge/serving/model-repo` mount to the `edge-endpoint` container in `edge-deployment.yaml`. The PVC was already mounted in that container at `/opt/groundlight/edge/sqlite`, but `MODEL_REPOSITORY_PATH` in `app/core/file_paths.py` resolves to `/opt/groundlight/edge/serving/model-repo`, where the `status-monitor` container already mounts it. Without this mount the new stamping code silently finds nothing.
- Additive and backward-compatible — no SDK changes, no `Posicheck` schema changes; older edges continue to omit the keys.

## Test plan

- [x] Unit: `pytest test/edge_inference/test_edge_inference_manager.py -v` — 7/7 pass (4 existing + 3 new: with-OODD, without-OODD, missing-model-id-nonfatal).
- [x] G4 integration (built `edge-endpoint:dev` via `deploy/bin/build-local-edge-endpoint-image.sh`, deployed via `make helm-install HELM_ARGS="--set edgeEndpointTag=dev ..."`):
  - [x] Happy path — fired 5 IQs against `det_2upHNr7q0U2xaX8atrVrqKmQ1Mb` (`edge_answers_with_escalation`). All 5 escalated with `edge_result.mlb_key` and `edge_result.oodd_mlb_key` matching live `model_id.txt` in the pod (primary v1, OODD v2).
  - [x] Negative path — moved primary `model_id.txt` aside; IQ still returned a valid result, `mlb_key` was cleanly absent, `oodd_mlb_key` remained present, no warnings in logs.
  - [x] Verified running image carries the new code: `kubectl exec ... grep 'Stamp the currently-loaded MLB' edge_inference.py`.

### DIFF
--- a/deploy/helm/groundlight-edge-endpoint/templates/edge-deployment.yaml
+++ b/deploy/helm/groundlight-edge-endpoint/templates/edge-deployment.yaml
@@ -207,6 +207,10 @@ spec:
           mountPath: /opt/groundlight/edge/sqlite
         - name: edge-endpoint-persistent-volume
           mountPath: /opt/groundlight/edge/config
+        # Needed by run_inference() to read {detector}/{primary,oodd}/{version}/model_id.txt
+        # and stamp the MLB KSUID into edge_result for cloud-side debugging.
+        - name: edge-endpoint-persistent-volume
+          mountPath: /opt/groundlight/edge/serving/model-repo
         - name: device-info-volume
           mountPath: /opt/groundlight/device
         - name: escalation-queue-volume
--- a/app/api/routes/image_queries.py
+++ b/app/api/routes/image_queries.py
@@ -197,6 +197,8 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
                 patience_time=patience_time,
                 rois=results["rois"],
                 text=results["text"],
+                mlb_key=results.get("mlb_key"),
+                oodd_mlb_key=results.get("oodd_mlb_key"),
             )
 
             # Skip cloud operations if escalation is disabled
--- a/app/core/edge_inference.py
+++ b/app/core/edge_inference.py
@@ -324,6 +324,28 @@ class EdgeInferenceManager:
 
         output_dict = get_inference_result(response, oodd_response, mode)
 
+        # Stamp the currently-loaded MLB KSUIDs into the result so that
+        # edge_result.mlb_key (and oodd_mlb_key when applicable) is persisted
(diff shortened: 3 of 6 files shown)


deploy/helm/groundlight-edge-endpoint/templates/edge-deployment.yaml
```diff
@@ -207,6 +207,10 @@ spec:
           mountPath: /opt/groundlight/edge/sqlite
         - name: edge-endpoint-persistent-volume
           mountPath: /opt/groundlight/edge/config
+        # Needed by run_inference() to read {detector}/{primary,oodd}/{version}/model_id.txt
+        # and stamp the MLB KSUID into edge_result for cloud-side debugging.
+        - name: edge-endpoint-persistent-volume
+          mountPath: /opt/groundlight/edge/serving/model-repo
```
honeytung, line 213: We need to mount model-repo for the edge-endpoint container to access primary and OODD inference KSUID.

timmarkhuff (approved):
Looks good. I think if we want to support this long-term, we should properly model it in Pydantic, instead of just putting stuff into the metadata.

Since this is for debugging an urgent issue, I think the expedient path is fine.

I did notice that you can only observe the mlb ids if: 1) your image query gets escalated to the cloud and 2) you re-fetch the image query from the cloud. The result that comes back initially from the edge does not have the metadata. Based on how we plan to use the data, that might be fine. However, it is a little confusing, and would make it impossible for users to see which mlb was used for local inference image queries. I asked Claude to mock out a change so this data is always returned, and it seemed pretty trivial.

Nit: the `run_inference` docstring is now stale -- still claims it returns `score` / `probability` /
`label` keys (it doesn't, and never did) and doesn't mention the new `mlb_key` / `oodd_mlb_key`. Was
already inaccurate pre-PR but worth a one-line touch since this PR is editing that function.

PR #384, opened 2026-04-22
Add more traces to Edge Endpoint profiling

timmarkhuff:
This PR closes remaining profiling trace gaps in request handling by adding missing `@trace_span` coverage and introducing runtime instrumentation around `run_in_threadpool` so dependency/background-task wait + execution time is visible in waterfalls.

It also annotates root request traces with `response_sent_ms`, making it easier to separate user-visible latency from post-response background work.

## Context from prior tracing PRs
- Builds on [#372](https://github.com/groundlight/edge-endpoint/pull/372), which introduced the core profiling/tracing system.
- Follows [#380](https://github.com/groundlight/edge-endpoint/pull/380), which expanded trace visibility and added the Marimo dashboard.
- Complements [#386](https://github.com/groundlight/edge-endpoint/pull/386), which improved dashboard interpretation for merged cross-process traces.

## What this adds
- Additional tracing on key request-path functions (`post_image_query`, `validate_content_type`, `get_app_state`, `get_groundlight_sdk_instance`, `run_inference`, and others).
- `install_threadpool_tracing()` monkey-patching of known `run_in_threadpool` bindings to expose sync-dispatch overhead as parent spans (including threadpool queue wait plus any event-loop resume delay around execution).
- Middleware capture of `response_sent_ms` to mark when the client response is fully flushed.
- Dashboard color map + profiling README updates for the newly visible spans.


app/profiling/instrumentation.py
```diff
@@ -0,0 +1,62 @@
...
+            return await original(func, *args, **kwargs)
+        finally:
+            tracer.end_span(span)
+            _current_span.reset(token)
+
+    setattr(traced, _PATCHED_MARKER, True)
+    return traced
+
+
+def install_threadpool_tracing() -> None:
```
f-wright, line 52: Did you run benchmarking to verify these threadpool changes didn't add latency?

app/profiling/README.md
```diff
@@ -18,6 +18,49 @@ ENABLE_PROFILING=true
 
 When enabled, the profiling middleware creates a trace per request. Functions decorated with `@trace_span` automatically create child spans. Traces are written as JSONL to `/opt/groundlight/device/edge-profiling/` with 5-minute file rotation and 24-hour automatic cleanup.
 
+## Traced Spans
+
+```
+request                                          <- middleware (full request lifecycle)
```
f-wright: We talked before about keeping this list out of the README, I think it's probably best if we just reference the code on this. Let me know what you think

app/main.py
```diff
@@ -28,6 +29,7 @@
 
 app = FastAPI(title="edge-endpoint")
 if PROFILING_ENABLED:
+    install_threadpool_tracing()
```
f-wright, line 37: Claude had some concerns here that I think are worth considering

> Stale run_in_threadpool bindings (medium)
> 
>   app/main.py:31 calls install_threadpool_tracing() after from app.api.api import api_router has already executed at module-load time. The patcher rebinds three known references:
> 
>   starlette.concurrency.run_in_threadpool = traced
>   fastapi.dependencies.utils.run_in_threadpool = traced
>   starlette.background.run_in_threadpool = traced
> 
>   But any third-party or first-party module that did from starlette.concurrency import run_in_threadpool before line 31 ran will hold a stale reference to the original. A grep of the repo is clean today, but this is a footgun for future code.
>    Two options:
> 
>   - Patch in-place: starlette.concurrency.run_in_threadpool.__code__ = traced.__code__ won't work since it's async def. Better — wrap by replacing the function object early enough, or
>   - Move the install above any router imports (e.g. into app/profiling/__init__.py import-time, gated on PROFILING_ENABLED), so it runs before any module aliases the symbol.
> 
>   At minimum, leave a comment in main.py warning that this must run before any future from starlette.concurrency import run_in_threadpool.

app/profiling/instrumentation.py
```diff
@@ -0,0 +1,62 @@
...
+    """Monkey-patch every known binding of run_in_threadpool to record the wait+execute span.
+
+    Idempotent: a second call is a no-op. Safe to call before the first request is served.
+    """
+    if getattr(starlette.concurrency.run_in_threadpool, _PATCHED_MARKER, False):
+        return
+    traced = _make_traced(starlette.concurrency.run_in_threadpool)
```
f-wright, line 59: I think it'd also be good to add some tests on this. Specifically, claude suggested a regression test like the one below, but it would also be worth testing the standard usage

```
def test_all_known_bindings_are_patched(self):
          """Regression test: fastapi/starlette.background must point at the patched fn.

          If a future Starlette/FastAPI release moves the import, this test catches it
          before production silently loses spans for sync deps / BackgroundTasks.
          """
          install_threadpool_tracing()
          patched = starlette.concurrency.run_in_threadpool
          assert fastapi.dependencies.utils.run_in_threadpool is patched
          assert starlette.background.run_in_threadpool is patched
```

app/profiling/instrumentation.py
```diff
@@ -0,0 +1,62 @@
...
+    """Wrap an async run_in_threadpool implementation so wait+execute time becomes a span."""
+
+    @functools.wraps(original)
+    async def traced(func: Callable[..., Any], *args: Any, **kwargs: Any) -> Any:
+        tracer = _current_tracer.get()
+        if tracer is None:
+            return await original(func, *args, **kwargs)
+        current = _current_span.get()
+        parent_id = current.span_id if current else None
+        name = f"run_in_threadpool[{getattr(func, '__name__', 'fn')}]"
```
f-wright, line 39: This created some overly long span names which makes the latency distribution by span chart render weirdly, and very difficult to read. Claude suggested flipping the chart to be vertical, and shortening this name or aliasing it for that chart. I think we should do something like that, so we can still use the chart.

f-wright (commented):
Left some comments that I think are worth considering, but overall this looks good! Very helpful to be able to see exactly where the missing time is going.

app/profiling/README.md
timmarkhuff: Agreed, I removed the tree and replaced it with a short note pointing at the source.

app/main.py
timmarkhuff: Added a comment above the call in main.py warning that if anyone ever adds from starlette.concurrency import run_in_threadpool to a local module, that import needs to come after this line. Otherwise the local binding won't be patched and those spans will silently disappear.

app/profiling/instrumentation.py
timmarkhuff: Added test/profiling/test_instrumentation.py, which includes the binding regression test you suggested plus no-op path, span creation, parent/child linking, and exception cleanup.

app/profiling/instrumentation.py
timmarkhuff: Flipped the Latency Distribution chart to horizontal boxes (span names on the y-axis, duration on x). Long names render cleanly now.

app/profiling/instrumentation.py
timmarkhuff: I just ran some benchmarking. See below. 

With this PR's changes, and profiling enabled:
image

Without this PR's changes, and profiling enabled:
image

Without this PR's changes, and profiling disabled:
image

They seem roughly the same. It's hard to tell. What do you think?

f-wright: Seems fine to me, it'd be nice to get more precise numbers from benchmarking in the future though

f-wright (approved):

app/profiling/instrumentation.py
timmarkhuff: The scripts do provide some numbers is a json file. I just didn't include them here

PR #386, opened 2026-04-27
Improve profiling dashboard plots and merge cross-process trace spans

f-wright:
## Summary
  Improvements to the Marimo profiling dashboard, focused on making cross-process traces actually readable and tightening a few plot-side rough edges.

  ## Changes

  **Cross-process trace merging** (`data_loader.py`)
  - New `merge_traces_by_id()` unifies records that share a `trace_id`. The edge endpoint and inference server each emit their own record per request, so without merging the waterfall only ever shows half the tree. Spans are deduped by `span_id`; metadata comes from the most "informative" record (real `detector_id` > `"unknown"` > empty, then span count).
  - The dashboard now loads traces *unfiltered* and applies the detector filter post-merge, because the inference-side record carries an empty `detector_id` and would otherwise be dropped before merge.

  **Waterfall hierarchy** (`dashboard.py`)
  - Spans are ordered by depth-first pre-order over `parent_span_id` with depth-indented y-axis labels, so children sit directly under their parent instead of getting reordered by start time.
  - Each span gets its own row (indexed by integer y-position) so spans that share a name across primary/OODD branches don't collide.
  - Descendants of `_submit_primary_inference` / `_submit_oodd_inference` inherit their wrapper's color, making it visually obvious which branch a given inference-pod span ran under.

  **Histogram UX**
  - Replaced the multi-subplot grid (unreadable when spans had wildly different timescales — 2ms vs 300ms) with a span-selector dropdown plus a single histogram. Default is `request`.

  **Latency-over-time scatter**
  - Now opt-in via a multiselect (defaults to the two inference wrappers). Plotting every span exceeded marimo's output payload size on busy traces.
  - Switched to `Scattergl` for perf, and unmapped spans cycle through a qualitative palette (hoisted to module constants) instead of all collapsing to the gray fallback.

  **Box plot** ordering switched from alphabetical to mean duration descending so the slowest spans surface first.

### DIFF
--- a/app/profiling/dashboard.py
+++ b/app/profiling/dashboard.py
@@ -35,10 +35,34 @@ def _():
     }
     FALLBACK_COLOR = "#B6B6B6"
 
+    # Qualitative palette used to auto-color spans that aren't in SPAN_COLORS.
+    AUTO_PALETTE = [
+        "#636EFA",
+        "#EF553B",
+        "#00CC96",
+        "#AB63FA",
+        "#FFA15A",
+        "#19D3F3",
+        "#FF6692",
+        "#B6E880",
+        "#FF97FF",
+        "#FECB52",
+        "#1F77B4",
+        "#FF7F0E",
+        "#2CA02C",
+        "#D62728",
+        "#9467BD",
+        "#8C564B",
+        "#E377C2",
+        "#7F7F7F",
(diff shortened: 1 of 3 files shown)


timmarkhuff (approved):
LGTM!

PR #387, opened 2026-04-27
Fix mount-s3.sh stale FUSE mount recovery

f-wright:
Reorder cleanup before `mkdir -p` so an orphaned FUSE endpoint (stat returns ENOTCONN) can be recovered automatically instead of crashing the sidecar under `set -e` before the cleanup branch is reached.


honeytung (approved):
LGTM! Thanks for looking into this!

PR #390, opened 2026-04-29
Use dog images in benchmarking scripts

timmarkhuff:
Some pipeline configs require COCO classes. Using non-COCO classes such as `circle` was causing training errors. 

This PR switches detector provisioning to use `dog` instead of `circle`, for both `COUNT` and `BOUNDING_BOX` detectors.

image-iq_3D0UWTGGq63Nr9bxCblhKhckLiu


load-testing/image_helpers.py
```diff
@@ -18,8 +21,81 @@ def generate_color_canvas(width: int, height: int, color: tuple[int, int, int])
     return np.full((height, width, 3), color, dtype=np.uint8)
 
 
+def _make_roi(label: str, x: int, y: int, dw: int, dh: int, image_width: int, image_height: int) -> ROI:
```
timmarkhuff, line 24: This is a better approach for creating ROIs, which doesn't depend on setting up a `Groundlight` client.

load-testing/memory_pressure_test.py
```diff
@@ -9,11 +9,10 @@
...
 import groundlight_helpers as glh
 import image_helpers as imgh
+from constants import OBJECT_DETECTION_CLASS_NAME
 
-SUPPORTED_DETECTOR_MODES = (
-    'BINARY',
-    'COUNT',
-)
+# Subset of SUPPORTED_DETECTOR_MODES that this script's load-generation path implements.
+MEMORY_PRESSURE_DETECTOR_MODES = ('BINARY', 'COUNT')
```
f-wright, line 15: Should this be swapped to bounding box mode?

load-testing/image_helpers.py
```diff
@@ -136,46 +165,35 @@ def generate_random_multiclass_image(
...
             image_width=image_width,
             image_height=image_height,
-            class_name=class_name,
             max_count=max_count,
         )
     elif detector_mode == 'BOUNDING_BOX':
-        config = detector.mode_configuration
-        class_name = config["class_name"]
-        max_num_bboxes = int(config.get("max_num_bboxes", 10))
+        max_num_bboxes = int(detector.mode_configuration.get("max_num_bboxes", 10))
```
f-wright: This has a fallback but max_count doesn't, I don't think either one of them need it

load-testing/image_helpers.py
```diff
@@ -18,8 +21,81 @@ def generate_color_canvas(width: int, height: int, color: tuple[int, int, int])
...
+    max_count: int = 10,
+) -> tuple[np.ndarray, int, list[ROI]]:
+    """Generate a white canvas with a random number of non-overlapping object images placed on it, with per-object ROIs."""
+    h, w = OBJECT_DETECTION_IMAGE.shape[:2]
+
+    count = random.randint(0, max_count)
+    canvas = generate_color_canvas(image_width, image_height, WHITE)
+
+    short_side = min(image_width, image_height)
+    min_size = max(1, int(short_side * 0.10))
```
f-wright, line 71: What do you think about making this configurable? I think it'd be useful for the future lens benchmarking where we might do something with the crops. Though maybe we'll be using separate images for that.

load-testing/image_helpers.py
```diff
@@ -18,8 +21,81 @@ def generate_color_canvas(width: int, height: int, color: tuple[int, int, int])
...
+
+        resized = cv2.resize(OBJECT_DETECTION_IMAGE, (dw, dh), interpolation=cv2.INTER_AREA)
+
+        for _ in range(50):
+            x = random.randint(0, max(0, image_width - dw))
+            y = random.randint(0, max(0, image_height - dh))
+            if not _boxes_overlap(placed, x, y, dw, dh):
+                break
+        else:
+            continue  # no non-overlapping position found; skip this object
```
f-wright: I think we should print some sort of error/warning message for this if we really want to skip objects, if someone sets their max_count too high this could systematically bias their benchmark towards fewer objects.

f-wright: It doesn't look like the prior implementation of this skipped objects. Are you more concerned about the overlap with dogs? It might still be fine and would prevent us from needing to skip objects.

f-wright (approved):
LGTM! Left a few comments for you to consider

load-testing/memory_pressure_test.py
timmarkhuff: Probably. Or better yet, we should should support all modes. I think that's an issue for another PR though. I didn't really want to touch this script in this PR, but Claude made this change as a part of a refactor, in order to not break this script. 

I would say if we have a use for this script again in the future, we can re-evaluate how this should work.

load-testing/image_helpers.py
timmarkhuff: Agreed. I'll remove the fallback.

load-testing/image_helpers.py
timmarkhuff: I'm thinking we would probably just use separate images for the crops. No need to actually crop the image. We'll see how that goes when we get there.

load-testing/image_helpers.py
timmarkhuff: The overlapping dogs just looked weird because of the background on the dog image. The logic to avoid overlapping is weird and potentially concerning. I removed that logic and images can now overlap. Near-white backgrounds are treated as transparent now.

PR #391, opened 2026-04-29
Add "Configuration" section to "Groundlight Edge Endpoint Status" page

timmarkhuff:
The "Groundlight Edge Endpoint Status" page now shows the configuration that the Edge Endpoint is using. The user can choose between yaml or json. yaml is default. 

image


honeytung (approved):
LGTM!

PR #392, opened 2026-04-30
Convert edge endpoint to uv with Helm compatibility

honeytung:
## Summary
- Convert the root edge-endpoint project from Poetry metadata and lockfile to uv-managed `pyproject.toml` and `uv.lock`
- Update Docker, Makefile, runtime scripts, CI workflows, integration scripts, deployment manifests, and docs from `poetry` commands to `uv` commands
- Add an image-owned `run_model_updater` command and update Helm/static k3s manifests to invoke it via `args`, so the chart no longer embeds the dependency manager for the model-updater sidecar
- Add a lightweight `poetry run` compatibility shim in the edge-endpoint image so old Helm charts can still run the new image while they are using `poetry run ...`
- Preserve main's split unit/integration CI changes and add `pytest-xdist` to the uv dev dependency group
- Harden the Helm/k3s install path by dumping pre-install token-validation hook diagnostics and retrying network-level `HTTP 000` credential fetch failures without retrying real HTTP auth/server responses


.github/workflows/pipeline.yaml
```diff
@@ -16,7 +16,8 @@ concurrency:
   cancel-in-progress: true
 env:
   PYTHON_VERSION: "3.11"
-  POETRY_VERSION: "2.1.3"
+  UV_VERSION: "0.8.4"
+  SDK_POETRY_VERSION: "2.1.3"
```
honeytung, line 20: We are still using `poetry` as the SDK uses that (we test the latest SDK in main too in CICD).

.github/workflows/auto-format.yaml
```diff
@@ -19,10 +19,10 @@ jobs:
         with:
           python-version: ${{ env.PYTHON_VERSION }}
         
-      - name: Install Poetry
-        uses: snok/install-poetry@v1.2.0
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
```
timmarkhuff, line 23: Out of curiosity, will this work when we migrate to GHEC? If not, is there an easy workaround?

timmarkhuff (approved):
LGTM

.github/workflows/auto-format.yaml
honeytung: I think it will work. A quick look into our GHEC workflow it looks like it is using `astral-sh/setup-uv@v7`.

timmarkhuff:
LGTM. I just did a backwards compatibility test, and it worked.

PR #394, opened 2026-05-01
Add CPU and GPU utilization to  `/status/resources.json` 

timmarkhuff:
## Summary

- Restructure `/status/resources.json` so CPU, RAM, and GPU metrics have explicit units and consistent attribution buckets.
- Poll inference pods via `/v2/gpu-usage`, placing VRAM under `gpu.vram_bytes` alongside GPU compute and memory-bandwidth utilization.
- Update the bundled status page, dev mock server, and resource metric tests for the breaking response shape.

## Response Structure Diff

```diff
 {
   "system": {
-    "ram": {
-      "used_bytes": 123,
-      "total_bytes": 456,
-      "loading_detectors_bytes": 0,
-      "edge_endpoint_bytes": 0,
-      "eviction_threshold_pct": 75
+    "cpu_utilization_pct": {
+      "total": 4.96,
+      "detectors": 2.1,
+      "loading_detectors": 0.0,
+      "edge_endpoint": 0.8,
+      "other": 2.06
     },
-    "vram": {
-      "used_bytes": 123,
-      "total_bytes": 456,
-      "loading_detectors_bytes": 0,
-      "edge_endpoint_bytes": 0,
-      "observed_gpus": [...]
+    "ram_bytes": {
+      "total": 16455819264,
+      "used": 10862436352,
+      "detectors": 4932083712,
+      "loading_detectors": 0,
+      "edge_endpoint": 1003692032,
+      "other": 4926660608,
+      "eviction_threshold_pct": 75
+    },
+    "gpu": {
+      "vram_bytes": {
+        "total": 16106127360,
+        "used": 2824011776,
+        "detectors": 2353004544,
+        "loading_detectors": 0,
+        "edge_endpoint": 0,
+        "other": 471007232
+      },
+      "compute_utilization_pct": {
+        "total": 0.0,
+        "detectors": 0.0,
+        "loading_detectors": 0.0,
+        "edge_endpoint": 0.0,
+        "other": 0.0
+      },
+      "memory_bandwidth_pct": {
+        "total": 0.0,
+        "detectors": 0.0,
+        "loading_detectors": 0.0,
+        "edge_endpoint": 0.0,
+        "other": 0.0
+      },
+      "devices": [
+        {
+          "index": 0,
+          "uuid": "GPU-...",
+          "name": "Tesla T4",
+          "vram_bytes": {
+            "total": 16106127360,
+            "used": 2824011776,
+            "free": 13282115584
+          },
+          "compute_utilization_pct": 0.0,
+          "memory_bandwidth_pct": 0.0
+        }
+      ]
     }
   },
   "detectors": [
     {
       "detector_id": "det_...",
-      "ram": { "primary_bytes": 1, "oodd_bytes": 2, "total_bytes": 3 },
-      "vram": { "primary_bytes": 1, "oodd_bytes": 2, "total_bytes": 3 }
+      "cpu_utilization_pct": { "primary": 1.0, "oodd": 0.8, "total": 1.8 },
+      "ram_bytes": { "primary": 1, "oodd": 2, "total": 3 },
+      "gpu": {
+        "vram_bytes": { "primary": 1, "oodd": 2, "total": 3 },
+        "compute_utilization_pct": { "primary": 0.0, "oodd": 0.0, "total": 0.0 },
+        "memory_bandwidth_pct": { "primary": 0.0, "oodd": 0.0, "total": 0.0 }
+      }
     }
   ]
 }
```

### DIFF
--- a/app/metrics/resource_metrics.py
+++ b/app/metrics/resource_metrics.py
@@ -120,12 +120,12 @@ class ResourceMetricsCollector:
         pod_list = v1.list_namespaced_pod(namespace=namespace)
         inference_pods = _find_inference_pods(pod_list)
         node_resources = _get_node_resources(v1)
-        ram_by_pod = _get_pod_ram_metrics(namespace)
+        pod_resources = _get_pod_resource_metrics(namespace)
 
         if inference_pods:
             active_pods = _pick_active_pods(inference_pods)
             gpu_responses = _query_all_pod_gpus(inference_pods)
-            observed_gpus, gpu_devices, total_vram, used_vram, gpu_compute_pct, gpu_memory_bw_pct = _build_gpu_summary(
+            _, gpu_devices, total_vram, used_vram, gpu_compute_pct, gpu_memory_bw_pct = _build_gpu_summary(
                 gpu_responses
             )
             (
@@ -134,17 +134,20 @@ class ResourceMetricsCollector:
                 loading_ram,
                 loading_gpu_compute_pct,
                 loading_gpu_memory_bw_pct,
+                loading_cpu_millicores,
             ) = _attribute_detector_resources(
                 inference_pods,
                 active_pods,
                 gpu_responses,
-                ram_by_pod,
+                pod_resources,
+                node_resources["cpu"]["total_millicores"],
             )
         else:
-            observed_gpus, gpu_devices, total_vram, used_vram = [], [], 0, 0
+            gpu_devices, total_vram, used_vram = [], 0, 0
             gpu_compute_pct, gpu_memory_bw_pct = 0.0, 0.0
             detectors, loading_vram, loading_ram = [], 0, 0
             loading_gpu_compute_pct, loading_gpu_memory_bw_pct = 0.0, 0.0
+            loading_cpu_millicores = 0.0
 
         # Anything in the namespace that isn't a recognised Running inference
         # pod counts as "edge-endpoint platform overhead": the edge-endpoint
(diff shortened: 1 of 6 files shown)


honeytung:
Tested PR end-to-end on a G4 (T4 GPU, edge-endpoint at `:dev`, `gl-edge-inference:pre-release`). Notes below.

## Validated

- **Schema match.** `/v2/gpu-usage` on `:pre-release` returns `schema_version: 2` with `process` block + `devices` array (`vram_total_bytes`/`vram_used_bytes`/`vram_free_bytes`/`compute_utilization_pct`/`memory_bandwidth_pct`). Matches the consumer in `app/metrics/resource_metrics.py` exactly.
- **`/status/resources.json` shape** is exactly what the PR description shows. Frontend, mock, tests are coherent with the new shape.
- **Non-zero CPU/GPU/mem-bw values reach the dashboard payload** under inference load.

## Known issues with per-process GPU utilization (deferred)

The `compute_utilization_pct` and `memory_bandwidth_pct` fields are populated end-to-end, but per-process attribution isn't yet reliable enough to drive a dashboard breakdown. Fine to keep in the schema as a placeholder for now — a more robust fix likely belongs upstream in `/v2/gpu-usage` (atomic per-device + per-process sampling, proper PID handling, per-process `gpu_index`). Issues observed during testing:

- **Sum of detector utilizations can exceed system total.** On a 1×T4 with one detector under load, samples like `total=61`, `detectors=64` are routine. `nvmlDeviceGetUtilizationRates` and `nvmlDeviceGetProcessUtilization` sample on independent ring buffers, so per-process sums can transiently exceed per-device readings.
- **Likely systematic under-reporting from container PID mismatch.** NVML reports samples by **host PID**. If the inference container filters "my process" using `os.getpid()` from inside its PID namespace, the filter matches nothing because the sample carries the host PID. Fixes inference-side: read `/proc/self/status` → `NSpid` to get the host PID, sum *all* process samples (correct on a single-tenant inference pod), or run with `hostPID: true`.
- **Multi-GPU scale mismatch is structural, not just measurement skew.** System rollup is `avg(per-device)` (always 0–100% regardless of N), while `detectors` is `sum(per-process across all GPUs)`. On a 2-GPU node with both at 80%, you'd legitimately get `total=80` but `detectors=160` — both factually correct but on different scales. No single-number rollup choice (avg, max, sum-clamped) reconciles cleanly without a per-device breakdown.
- **No `gpu_index` on detector entries.** The `devices[]` array exposes per-device readings, but there's no way to attribute a detector to a specific GPU. The cleanest UX is probably one stacked bar per device with each detector entry carrying a `gpu_index` field — but that requires schema additions on both `/v2/gpu-usage` and `/status/resources.json`.

## Minor nits (non-blocking, not addressed)

- `_K8S_CPU_SUFFIX_TO_MILLICORES.get(suffix, Decimal("1000"))` fallback is unreachable (regex restricts suffix to `[num]`).
- `_blank_resource()` returns `int 0` for `total` — reused for percentage fields too. Inconsequential.
- A few test docstrings in `test_resource_metrics.py` still reference old keys (`oodd_bytes=None`, `primary or oodd slot`).

## Side note (unrelated to this PR)

While testing, found that the canonical curl-based inference call from the G4 testing playbook (`-H "Content-Type: image/jpeg" --data-binary @file` over HTTPS:30143) hits the inference server with corrupted bytes when curl uses chunked transfer encoding. Workarounds: explicit `Content-Length`, `--http1.0`, or HTTP port 30101. Not introduced by this PR — likely a pre-existing nginx/TLS interaction. Worth a separate investigation.

honeytung (approved):
LGTM, added a comment about potential problems but since we did not use those metrics at the moment we can revisit those later. This just sets a standard format that we can use.

PR #395, opened 2026-05-04
Enable compilation and prevent benchmarking cache hits

f-wright:
- Adds `edgeCompileEnabled` Helm value (default false) that sets `EDGE_COMPILE_ENABLED` on inference pods, allowing `torch.compile` to be toggled per deployment without modifying the chart.
  - Bumps the load-testing binary image generator's timestamp overlay from second resolution (%Y-%m-%d %H:%M:%S) to millisecond resolution (%H:%M:%S.%f truncated). Previously, all queries within the same wall-clock second produced
  byte-identical images, which collapsed to a single cache key in generic-cached-* pipelines and hid the encoder workload behind the feature cache.

### DIFF
--- a/load-testing/image_helpers.py
+++ b/load-testing/image_helpers.py
@@ -106,7 +106,7 @@ def generate_random_binary_image(
         text_color = BLACK
         label = "NO"
 
-    timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    timestamp = datetime.now().strftime("%H:%M:%S.%f")[:-3]
     cv2.putText(image, timestamp, (50, 100), cv2.FONT_HERSHEY_SIMPLEX, 1, text_color, 2)
 
     return image, label, None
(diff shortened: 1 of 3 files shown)


load-testing/image_helpers.py
```diff
@@ -106,7 +106,7 @@ def generate_random_binary_image(
         text_color = BLACK
         label = "NO"
 
-    timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    timestamp = datetime.now().strftime("%H:%M:%S.%f")[:-3]
```
timmarkhuff, line 109: small images (e.g. 100 x 100) will have the unique part cut off

I think we should either scale the text to fit within the canvas (cv2 has some functions for this) and/or add a random hash to the image to ensure uniqueness across function calls.

timmarkhuff (approved):

PR #397, opened 2026-05-06
Fix NameError on oodd_deployment_name in minimal-image model update path

f-wright:
When separate_oodd_inference is false, oodd_deployment_name was never bound but was referenced unconditionally in rollout log/error messages, causing an error during in-place model updates on minimal-image clusters.


timmarkhuff (approved):
LGTM

PR #398, opened 2026-05-06
Source load-test system utilization from /status/resources.json

honeytung:
## Summary
- Replace `psutil` + `GPUtil` sampling in `SystemMonitor` with HTTP polling of the Edge Endpoint's `/status/resources.json` (added in #394). Sampling on the load-driver host is wrong when the driver runs on a different machine than the system under test.
- Only system-level totals (`cpu_utilization_pct.total`, `ram_bytes`, `gpu.compute_utilization_pct.total`, `gpu.vram_bytes`) are consumed. Per-detector / `loading_detectors` / `other` buckets are intentionally ignored — see [#394 review discussion](https://github.com/groundlight/edge-endpoint/pull/394#issuecomment-4382862153) for the per-process attribution caveats.
- Preserves the JSONL schema (`cpu_percent`, `memory_percent`, `gpu_utilization`, `vram_utilization`) so `parse_load_test_logs.py` and downstream plots are unchanged.
- New `--system-sample-interval-seconds` flag (default 5s) on `multiple_client_throughput_test.py`. CPU/RAM are bounded by Kubernetes Metrics Server cadence (~15s) so adjacent samples will repeat CPU/RAM while GPU swings — this is documented in the README.
- Drops `gputil` and `psutil` from `load-testing/pyproject.toml` (no other consumers in `load-testing/`); regenerated `uv.lock`.

## Implementation notes
- `_get_resources(gl)` mirrors the existing `_get_status_metrics(gl)` helper in `groundlight_helpers.py`. `call_api` gained an optional `timeout` kwarg (default `None` preserves existing behavior); `_get_resources` passes `timeout=5.0` so a slow/hung Edge Endpoint during a load test can't stall the monitor.
- The monitor process constructs its own `ExperimentalApi()` from the `GROUNDLIGHT_ENDPOINT` env var (matching the existing pattern in `send_image_requests`) — avoids pickling a client across the `multiprocessing.Process` boundary.
- Resource-fetch errors are caught, logged at most once every 30s, and the tick is skipped — the load test keeps running even if the Edge Endpoint is momentarily unhealthy.

## Test plan
- [x] Mock `glh._get_resources` and drive `_monitor_loop` for a few ticks; verify JSONL shape matches the existing parser expectations.
- [x] Run `multiple_client_throughput_test.py BINARY --max-clients 4 --time-between-ramp 30` against a live G4 Edge Endpoint built from this branch; confirm `load_test.log` contains `event: "cpu"` and `event: "gpu"` lines at ~5s cadence.
- [x] Re-run `parse_load_test_logs.py` and confirm `throughput_and_system_utilization_over_time.png` and `time_vs_latency.png` render unchanged.
- [x] Cross-check the GPU line against `nvtop` and CPU against `kubectl top node` (allowing for the ~15s Metrics Server window).
- [x] Tail nginx access log on the Edge Endpoint; confirm ~120 `/status/resources.json` requests over a 10-minute test at 5s cadence.

### DIFF
--- a/load-testing/README.md
+++ b/load-testing/README.md
@@ -52,7 +52,12 @@ If `--edge-pipeline-config` is omitted, the detector's current/default edge pipe
 ### Multiple Client Throughput Test
 
 #### Purpose
-Tests the Edge Endpoint's ability to handle concurrent client load by spawning multiple client processes that ramp up gradually from 1 to N workers, measuring throughput and latency over time. During the test, host system utilization (CPU/GPU/RAM/VRAM) is also sampled and included in the generated artifacts.
+Tests the Edge Endpoint's ability to handle concurrent client load by spawning multiple client processes that ramp up gradually from 1 to N workers, measuring throughput and latency over time. During the test, system utilization (CPU/GPU/RAM/VRAM) is sampled by polling the Edge Endpoint's `/status/resources.json` and included in the generated artifacts.
+
+#### System utilization sampling
+Sampling cadence is controlled by `--system-sample-interval-seconds` (default 5s). Two notes:
+* Only system-level totals (`system.cpu_utilization_pct.total`, `system.ram_bytes`, `system.gpu.compute_utilization_pct.total`, `system.gpu.vram_bytes`) are recorded. Per-detector / `loading_detectors` / `other` buckets are intentionally ignored.
+* GPU/VRAM is fresh-on-call (~1s NVML ring buffer). CPU/RAM is bounded by the Kubernetes Metrics Server cadence (~15s default refresh). Adjacent samples will show GPU swinging while CPU/RAM stays flat — that is expected, not a bug.
 
 #### Usage
 ```
--- a/load-testing/groundlight_helpers.py
+++ b/load-testing/groundlight_helpers.py
@@ -46,14 +46,14 @@ class APIError(Exception):
     """
     pass
 
-def call_api(url: str, params: dict) -> dict:
+def call_api(url: str, params: dict, timeout: float | None = None) -> dict:
     """Perform a GET request with API token and return decoded JSON or raise APIError."""
 
     headers = {
         "X-API-Token": os.environ.get('GROUNDLIGHT_API_TOKEN')
     }
-    
-    response = requests.get(url, params=params, headers=headers)
+
+    response = requests.get(url, params=params, headers=headers, timeout=timeout)
     if response.status_code == 200:
         response_content = response.content.decode('utf-8')
         return json.loads(response_content)
@@ -130,6 +130,12 @@ def _get_status_metrics(gl: ExperimentalApi) -> dict:
     base = gl.endpoint.replace('/device-api', '')
     url = base + '/status/metrics.json'
     return call_api(url, {})
+
+def _get_resources(gl: ExperimentalApi, timeout: float = 5.0) -> dict:
+    """Retrieve the resource utilization JSON from the Edge Endpoint's /status/resources.json."""
+    base = gl.endpoint.replace('/device-api', '')
+    url = base + '/status/resources.json'
+    return call_api(url, {}, timeout=timeout)
 
 def get_container_images_map(gl: ExperimentalApi) -> dict[str, dict[str, str]]:
     """Return a map of pod -> {container: image_id} from edge status metrics."""
--- a/load-testing/multiple_client_throughput_test.py
+++ b/load-testing/multiple_client_throughput_test.py
@@ -204,6 +204,7 @@ def main(  # noqa: PLR0913
     image_height: int = 480,
     edge_pipeline_config: str | None = None,
     n: int | None = None,
+    system_sample_interval_seconds: float = 5.0,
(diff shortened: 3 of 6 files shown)


load-testing/groundlight_helpers.py
```diff
@@ -131,6 +131,12 @@ def _get_status_metrics(gl: ExperimentalApi) -> dict:
     url = base + '/status/metrics.json'
     return call_api(url, {})
 
+def _get_resources(gl: ExperimentalApi, timeout: float = 5.0) -> dict:
+    """Retrieve the resource utilization JSON from the Edge Endpoint's /status/resources.json."""
+    base = gl.endpoint.replace('/device-api', '')
+    url = base + '/status/resources.json'
+    return call_api(url, {}, timeout=timeout)
```
honeytung, line 138: I was debating if we want to fetch the `ENDPOINT` through the SDK object because it looks weird but it is the convention that we have been doing so I will keep it for now.

load-testing/multiple_client_throughput_test.py
```diff
@@ -303,5 +304,13 @@ def main(  # noqa: PLR0913
     parser.add_argument("--image-width", type=int, default=640)
     parser.add_argument("--image-height", type=int, default=480)
     parser.add_argument("--edge-pipeline-config", type=str, default=None, help="Edge pipeline configuration name.")
+    parser.add_argument(
+        "--system-sample-interval-seconds", type=float, default=5.0,
+        help=(
+            "How often to poll the Edge Endpoint's /status/resources.json for system utilization. "
+            "CPU/RAM are bounded by Kubernetes Metrics Server cadence (~15s); values below 15s give "
+            "repeated CPU/RAM samples but fresher GPU/VRAM samples."
```
honeytung, line 318: Note that the 15s poll rate is a fixed rate set by the built-in k3s metrics server. If we want to change it we will need to disable the built-in one and use a custom manifest which is not ideal.

load-testing/README.md
```diff
@@ -52,7 +52,12 @@ If `--edge-pipeline-config` is omitted, the detector's current/default edge pipe
 ### Multiple Client Throughput Test
 
 #### Purpose
-Tests the Edge Endpoint's ability to handle concurrent client load by spawning multiple client processes that ramp up gradually from 1 to N workers, measuring throughput and latency over time. During the test, host system utilization (CPU/GPU/RAM/VRAM) is also sampled and included in the generated artifacts.
+Tests the Edge Endpoint's ability to handle concurrent client load by spawning multiple client processes that ramp up gradually from 1 to N workers, measuring throughput and latency over time. During the test, system utilization (CPU/GPU/RAM/VRAM) is sampled by polling the Edge Endpoint's `/status/resources.json` and included in the generated artifacts.
+
+#### System utilization sampling
+Sampling cadence is controlled by `--system-sample-interval-seconds` (default 5s). Two notes:
+* Only system-level totals (`system.cpu_utilization_pct.total`, `system.ram_bytes`, `system.gpu.compute_utilization_pct.total`, `system.gpu.vram_bytes`) are recorded. Per-detector / `loading_detectors` / `other` buckets are intentionally ignored.
```
honeytung, line 59: Those individual detector metrics are intentionally excluded as those are not in the original load testing. We also found there might be a problem with our individual GPU utilization that it is not reporting metrics accurately (see comments in #394). We can revisit this after we figure out how to improve those.

f-wright: This is interesting, I wasn't aware the individual detector metrics were incorrect. Is there a plan for resolving that yet?

load-testing/multiple_client_throughput_test.py
f-wright: Would it make sense to use separate rates for the CPU and GPU samples then, to avoid potentially misleading repeated samples? I think this way is fine too though, and I guess we'd need to rework things to sample separately if we made that change

load-testing/system_helpers.py
```diff
@@ -1,18 +1,32 @@
...
 
 
 class SystemMonitor:
-    """Monitors CPU and GPU utilization in a background process and logs results."""
+    """Polls the Edge Endpoint's /status/resources.json in a background process and logs results.
+
+    Emits one `event: "cpu"` and one `event: "gpu"` JSONL record per tick, matching
+    the schema parse_load_test_logs.py consumes. Only system-level totals are recorded;
+    per-detector / loading_detectors / other buckets are intentionally ignored
+    (NVML windowing skew, container PID mismatch — see PR #394 review discussion).
```
f-wright, line 23: Are we going to fix this soon? Seems a bit odd to have this in the docstring, but if it's just a TODO that will be removed soon I suppose that's ok

f-wright (approved):
Left a few comments, but LGTM!

load-testing/multiple_client_throughput_test.py
honeytung: Yea, we can talk about this more, but it might make sense to make those metrics be reporting at the same frequency since CPU and GPU utilization are both important and time sensitive metrics to look at.

PR #406, opened 2026-05-13
Lens Benchmarking V1

honeytung:
## Summary

A simplified, lens-shape-driven benchmark for the edge-endpoint that replaces per-frame composite/chain/crop logic with three fixed lens shapes and a clean `n`-sweep. Built next to (not on top of) the existing `app_benchmark` harness on `htung_axoncorp/edge-bench-impl` so both can be evaluated side-by-side.

- Three lens shapes:
  - `single_binary` — one binary inference per frame.
  - `single_bbox` — one bbox inference per frame; `n` sets `max_num_bboxes` (sweepable).
  - `bbox_to_binary` — one bbox call + `n` downstream binary calls per frame.
- Sweep dimension is a per-lens `n` list, zipped across lenses. Lenses run concurrently within each run; runs are sequential.
- Detectors are provisioned **once** with `max(lens.n)` and reused across all runs — eliminates 4–8× retraining overhead.
- Reuses existing load-testing helpers: `groundlight_helpers.provision_detector`, `image_helpers.generate_random_*`, `system_helpers.SystemMonitor`, `parse_load_test_logs._percentile`.

## What's in the output

- Single consolidated `summary.md` with environment block, ICMP ping baseline, cross-run overview table, combined cross-run plots, and per-run sections.
- Cross-run plots at `out_root/plots/`:
  - `system_utilization.png` — 2×2 grid (CPU%, GPU%, RAM GB, VRAM GB) on benchmark-wall-clock x-axis, with vertical lines labeled `Run i` at each run's `main_start_ts`.
  - `fps_{lens}_camera_{N}.png` per camera process, same x-axis, with vertical lines labeled `Run i (n=X)` using that lens's own `n` for the run.
- Per-run `run_NN/plots/` for drill-down.
- Per-run + top-level `summary.json` for machine consumption.

## Config example

```yaml
run:
  name: "sample-bench"
  edge_endpoint_url: "http://localhost:30101"
  detector_name_prefix: "bench"
  refuse_if_host_not_clean: true

global:
  image_size: [1920, 1080]
  target_fps: 5
  duration_seconds: 180
  warmup_seconds: 20

lenses:
  - {name: door_lens, type: single_binary, pipeline: ..., cameras: 1, image_size: [640, 480], target_fps: 10}
  - {name: person_lens, type: bbox_to_binary, bbox_pipeline: ..., binary_pipeline: ..., n: [2, 4, 6, 8], cameras: 1}
  - {name: object_lens, type: single_bbox, pipeline: ..., n: [1, 3, 5, 7], cameras: 1, target_fps: 0}

monitoring:
  sample_hz: 1.0
```

Per-lens `image_size` and `target_fps` overrides (target_fps: 0 = saturate). `n` lists must share length; harness zips across lenses to produce one run per index.

### Plots

fps_door_lens_camera_0
fps_object_lens_camera_0
fps_person_lens_camera_0
system_utilization

### Summary.md

Benchmark: sample-bench

- **Started**: 2026-05-13T23:04:00.903012+00:00
- **Edge endpoint**: `http://localhost:30101`
- **Network ping baseline**: loopback — skipped
- **Defaults**: image_size `[1920, 1080]`, target_fps `5.0`, duration `120s`, warmup `20s`

Overview

Per-lens FPS = mean across that lens's camera processes. Hit = `yes` if every camera hit ≥95% of its target.

| run | lens_n | door_lens FPS | door_lens Hit | object_lens FPS | object_lens Hit | person_lens FPS | person_lens Hit |
|---|---|---|---|---|---|---|---|
| 0 | `{'person_lens': 2, 'object_lens': 1}` | 10.0 | yes | 5.7 | — | 4.7 | no |
| 1 | `{'person_lens': 4, 'object_lens': 3}` | 10.0 | yes | 5.7 | — | 4.1 | no |
| 2 | `{'person_lens': 6, 'object_lens': 5}` | 10.0 | yes | 5.6 | — | 3.6 | no |
| 3 | `{'person_lens': 8, 'object_lens': 7}` | 10.0 | yes | 5.5 | — | 3.2 | no |

## Test plan

- [x] Config validation (positive + negative): unequal n-list lengths, duplicate names, bad lens type all raise `ConfigError`.
- [x] Per-lens `image_size` / `target_fps` overrides parse and apply.
- [x] Smoke test on G4 (`smoke.yaml`, 1 single_binary, 30s): 150 frames at 5 FPS, 0 errors.
- [x] 2-camera smoke on G4: per-camera FPS plots generated.
- [x] Multi-run sweep on G4 (`sweep_smoke.yaml`, n=[2,6]): detector reuse confirmed (one bbox + one binary trained, both reused across both runs); cross-run plots show `Run 0 (n=2)` / `Run 1 (n=6)` annotations correctly.
- [x] SystemMonitor fix: cpu/gpu events written; 2×2 system grid renders with absolute GB and total reference lines.
- [x] Network ping baseline: ICMP ping result in summary.md (loopback correctly skipped on G4).
- [x] Percentile sanity: p50 of [1,2] = 1.5; p95 of 1..20 = 19.05 (was 20 with the old formula).
- [x] Window enforcement: late-tail event with ts > main_end_ts is excluded.
- [x] Missing-camera detection: expected `(lens, camera)` row emitted even when no events; `hit_target=False`, flagged in summary.
- [ ] Verify on G4 with the five fixes in `d9ce63a` (sweep_smoke, includes the latency-percentile sanity check).


load-testing/app_benchmark/host_check.py
```diff
@@ -0,0 +1,64 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class HostNotCleanError(RuntimeError):
+    """Raised by ensure_host_clean when the edge already has detectors
+    configured and refuse_if_host_not_clean is True."""
+
+
+def ensure_host_clean(gl_edge: ExperimentalApi, *, allow: bool = False) -> None:
```
timmarkhuff, line 21: I'm not sure I understand the purpose of this function. Why does it matter if the host has other detectors configured? gl.edge.set_config resolves that.

I just tried to run the script on my machine and I got this error:
```
(.venv) thuff@AM-VJ75H3YMTG load-testing % uv run python -m app_benchmark app_benchmark/configs/simple_bbox_plus_binary.yaml
16:29:26 INFO    app_benchmark: output dir: benchmark-results/simple-bbox-binary-20260513T232926Z
16:29:26 INFO    app_benchmark: network ping baseline: loopback — skipped
16:29:26 ERROR   app_benchmark: host check failed: Edge has 1 pre-existing detector(s) configured: ['det_38CoIPd3ysjoMemNprqDSM9NzC5']. Wipe with: python -m app_benchmark.cleanup_edge --edge-endpoint <URL> --wipe. Or set run.refuse_if_host_not_clean: false in the YAML to proceed (results will skew).
```

It looks like I should run something like this to resolve the issue:
```bash
uv run --active python -m app_benchmark.cleanup_edge --edge-endpoint http://localhost:30101 --wipe
```

But why do I need to do that?

honeytung: I left that just as a sanity check to make sure the edge-endpoint is not running any inference from other apps/experiments. Even if we apply the config to delete the detectors not used, if someone is running some apps it may call edge-endpoint and spin up the inference pods again.

If you think that is not necessarily then we can remove that and only show a warning like:

```
WARNING: edge-endpoint has pre-configured detectors that indicates the edge-endpoint might be used for other applications.
The script will proceed but the results may be different if edge-endpoint is in use.
```

timmarkhuff:
If you look in https://github.com/groundlight/edge-endpoint/pull/373, you will see that I was working on a slightly different mechanism for provisioning detectors. Instead provisioning each detector sequentially, I did them all together (i.e. submit labels for all detectors, and then wait for all detectors to finish training). It's way faster if your script is going to be using many detectors, because the training can happen in parallel. 

Up to you whether it's worth it to add in this PR. If you're interested, look at the `provision_all` function in https://github.com/groundlight/edge-endpoint/pull/373.

load-testing/app_benchmark/README.md
```diff
@@ -0,0 +1,157 @@
+# app_benchmark — Lens-sweep benchmark for the edge-endpoint
+
+A simplified harness that runs **three fixed lens shapes** concurrently and
+**sweeps over the bbox `n` parameter** to characterize how throughput and
```
f-wright: Is the `n` parameter max_num_bboxes? should we just say that instead?

f-wright: Or is it the number of objects actually present in the image? Either way I think more specificity would be good here

load-testing/app_benchmark/README.md
```diff
@@ -0,0 +1,157 @@
...
+|---|---|---|---|
+| `single_binary` | 1 binary call | no | binary inference on a pre-encoded synthetic JPEG |
+| `single_bbox` | 1 bbox call | yes | bbox inference on a synthetic objects image bounded by `n` |
+| `bbox_to_binary` | 1 bbox + N binary | yes | one bbox call, then N binary calls reusing one tiny pre-encoded JPEG |
+
+For n-bearing lenses, the bbox detector is provisioned **once** with
+`max_num_bboxes = max(lens.n)` and reused across every run in the sweep.
+The per-run `n` only varies the worker's behavior:
+
+- the synthetic image's object-count bound (`generate_random_objects_image(max_count=n)`), and
```
f-wright: It would be nice to be able to control the actual number of objects in the images, not just the max.

load-testing/app_benchmark/README.md
```diff
@@ -0,0 +1,157 @@
...
+- for `bbox_to_binary`, the number of downstream binary calls issued per frame.
+
+All lenses pre-encode a small JPEG pool at startup and cycle per frame —
```
f-wright: Maybe we should call them "synthetic lenses" for clarity?

load-testing/app_benchmark/README.md
```diff
@@ -0,0 +1,157 @@
...
+Within each run, every lens runs its full `cameras` count in parallel as
+independent OS processes.
+
+**Detector lifecycle is one-shot, not per-run.** Each n-bearing lens
+provisions a single bbox detector with `max_num_bboxes = max(lens.n)`
+(the upper bound across the sweep) and trains it once before any run
+starts. The per-run `n` value only controls:
+
+- the synthetic image's object-count bound (`generate_random_objects_image(max_count=n)`), and
+- for `bbox_to_binary` lenses, the number of downstream binary calls issued per frame.
```
f-wright: Does `n` actually control this, or will it be based on the number of objects in the image?

load-testing/app_benchmark/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+The inference cost of the bbox model itself is essentially independent
```
f-wright: Have you verified this?

load-testing/app_benchmark/README.md
```diff
@@ -0,0 +1,157 @@
...
+of `max_num_bboxes` — the model processes the whole image regardless;
+only NMS post-processing depends on detected count, which is negligible
+relative to the convolutional cost.
+
+## Hard dependencies
+
+- Edge-endpoint includes PR #394 (`/status/resources.json` v2 shape).
+- `GROUNDLIGHT_API_TOKEN` exported in the shell.
+- Edge runs in Kubernetes (k3s) with x86 + ≥1 NVIDIA GPU.
```
f-wright: Why couldn't it be run on CPU? I don't really think we'll need to benchmark without the GPU, but it'd be nice to not prevent it

load-testing/app_benchmark/README.md
```diff
@@ -0,0 +1,157 @@
...
+- for `bbox_to_binary` lenses, the number of downstream binary calls issued per frame.
+
+The inference cost of the bbox model itself is essentially independent
+of `max_num_bboxes` — the model processes the whole image regardless;
+only NMS post-processing depends on detected count, which is negligible
+relative to the convolutional cost.
+
+## Hard dependencies
+
+- Edge-endpoint includes PR #394 (`/status/resources.json` v2 shape).
```
f-wright: I think we can take this out of the readme, this PR is going in after that one.

load-testing/app_benchmark/README.md
```diff
@@ -0,0 +1,157 @@
...
+  that lens's own `n` for the run.
+- Per-run sections with a per-camera table — `Frames`, `Errors`, `FPS`,
+  `Target`, `Hit`, `p50`, `p95` — and a ⚠ callout for any worker that
+  exited non-zero or any camera that produced no events.
+
+The FPS plots show frames-per-second on the left y-axis (blue), target FPS
+as a dashed orange line, and failed requests / sec on a right y-axis (red).
+For chained `bbox_to_binary` lenses, **frame** means one lens-loop iteration
+(one bbox call + N binary calls); the errors axis counts all failed
+**requests** across stages.
```
f-wright: Is a request here a single inference request, or a single full frame request (containing multiple inferences)?

load-testing/app_benchmark/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+The measurement window is fixed at `[main_start_ts, main_start_ts +
+duration)`. Events before `main_start_ts` (warmup) or at/after
+`main_end_ts` (any in-flight or grace-period requests) are excluded from
+the summary, so FPS reflects exactly `total_frames / duration`.
+
+## Cleanup utilities
+
+The benchmark cleans up its own detectors (cloud-side delete + edge-config
```
f-wright, line 128: Why do we need to clean them up? I don't think there's much harm to letting them linger in the cloud, especially if that would allow us to skip retraining and speed up the process of future benchmarks.

load-testing/.gitignore
```diff
@@ -0,0 +1,4 @@
+benchmark-results/
+load_tests/
+__pycache__/
```
f-wright: This and .venv are already covered recursively from the root gitignore, so we can remove them too

load-testing/.gitignore
```diff
@@ -0,0 +1,4 @@
+benchmark-results/
+load_tests/
```
f-wright: This is also in the root gitignore already, can we just move `benchmark-results` there? Also, maybe we should swap the dash for an underscore for consistency.

load-testing/system_helpers.py
```diff
@@ -92,6 +93,10 @@ def _monitor_loop(log_file: str, sample_interval: float, stop_event) -> None:
             vram_used = vram.get("used", 0)
             vram_percent = (vram_used / vram_total * 100) if vram_total else 0.0
 
+            ram_used_gb = ram_used / 1e9
```
f-wright: Can you make 1e9 a constant in the file to avoid repetition here? Or make a bytes to gb function that does the division

load-testing/app_benchmark/configs/example.yaml
```diff
@@ -0,0 +1,44 @@
...
+    n: [2, 4, 6, 8]
+    cameras: 1
+    # uses global image_size + target_fps
+
+  - name: object_lens
+    type: single_bbox
+    pipeline: bounding-boxes-step-rfdetr-primed
+    n: [1, 3, 5, 7]
+    cameras: 1
+    target_fps: 0            # saturate — go as fast as the edge can serve
```
f-wright, line 41: Instead of a 0 here, maybe it would make more sense to just leave this option unset when we want to go as fast as possible?

load-testing/app_benchmark/configs/example.yaml
```diff
@@ -0,0 +1,44 @@
...
+    image_size: [640, 480]   # smaller image for the binary stage
+    target_fps: 10           # binary is cheap — push it harder
+
+  - name: person_lens
+    type: bbox_to_binary
+    bbox_pipeline: bounding-boxes-step-rfdetr-primed
+    binary_pipeline: generic-cached-timm-efficientnetv2s-calibrated-mlp
+    n: [2, 4, 6, 8]
+    cameras: 1
+    # uses global image_size + target_fps
```
f-wright, line 34: Using the global target_fps here is a bit weird to me. I'd think that the global target FPS would be for the whole lens, so we'd want the individual component pipelines to run faster than that to reach that target FPS

load-testing/app_benchmark/configs/example.yaml
```diff
@@ -0,0 +1,44 @@
+run:
+  name: "sample-bench"
+  output_dir: "./benchmark-results/{name}-{ts}/"
+  edge_endpoint_url: "http://localhost:30101"
+  detector_name_prefix: "bench"
```
f-wright: Could we just use the name of the run here? I think there's a lot of configuration options and it'd be nice to simplify a bit

load-testing/app_benchmark/configs/example.yaml
```diff
@@ -0,0 +1,44 @@
+run:
+  name: "sample-bench"
+  output_dir: "./benchmark-results/{name}-{ts}/"
```
f-wright: Can this just be automatic, instead of configured?

load-testing/app_benchmark/configs/example.yaml
```diff
@@ -0,0 +1,44 @@
...
+  edge_endpoint_url: "http://localhost:30101"
+  detector_name_prefix: "bench"
+  refuse_if_host_not_clean: true
+
+global:
+  image_size: [1920, 1080]
+  target_fps: 5         # 0 = saturate (no pacing)
+  duration_seconds: 180
+  warmup_seconds: 20
```
f-wright: A comment on what the duration + warmup means would make this example a bit more instructive I think

load-testing/app_benchmark/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Lens benchmarking for testing edge-endpoint load capacity.
```
f-wright: I think we can delete this file, this info is all in the readme

load-testing/app_benchmark/cleanup_edge.py
```diff
@@ -0,0 +1,78 @@
...
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--edge-endpoint", required=True, help="Edge endpoint URL.")
+    parser.add_argument("--list", action="store_true", help="Print currently-loaded detectors and exit.")
+    parser.add_argument("--wipe", action="store_true", help="Push an empty EdgeEndpointConfig.")
+    parser.add_argument("--confirm", action="store_true", help="With --wipe, require interactive confirmation.")
+    args = parser.parse_args(argv)
+
+    logging.basicConfig(level=logging.INFO, format="%(message)s")
+    if not args.list and not args.wipe:
+        parser.error("must pass at least one of --list, --wipe")
```
f-wright: Could one of these just be the default, and the other a special case we can hit with the flag, instead of erroring here? I think it'd make senes to always list, and only wipe if that's passed.

load-testing/app_benchmark/cleanup_edge.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Returns:
+        Exit code: 0 success, 1 wipe failed or user aborted at confirm,
+        2 could not reach edge / bad usage.
+    """
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--edge-endpoint", required=True, help="Edge endpoint URL.")
+    parser.add_argument("--list", action="store_true", help="Print currently-loaded detectors and exit.")
+    parser.add_argument("--wipe", action="store_true", help="Push an empty EdgeEndpointConfig.")
+    parser.add_argument("--confirm", action="store_true", help="With --wipe, require interactive confirmation.")
```
f-wright: I think this should be the default value, and have a force flag if you want it to skip confirmation. I don't think it's useful to make the user pass an extra flag if they want to be asked again, that won't prevent accidental runs

load-testing/app_benchmark/cleanup_edge.py
```diff
@@ -0,0 +1,78 @@
...
+        logger.error("could not fetch current edge config: %s", exc)
+        return 2
+
+    det_entries = getattr(current, "detectors", None) or []
+    det_ids = [d.detector_id for d in det_entries if hasattr(d, "detector_id")]
+    logger.info("Edge has %d detector(s) configured:", len(det_ids))
+    for det_id in det_ids:
+        logger.info("  - %s", det_id)
+
+    if args.list and not args.wipe:
```
f-wright:
```suggestion
    if not args.wipe:
```
They can't both be true, so we don't need that check

load-testing/app_benchmark/cli.py
```diff
@@ -0,0 +1,395 @@
...
+    """Pick the `n` value at position `run_index` from every lens that
+    declares an `n` list. Lenses without `n` are absent from the result."""
+    out: dict[str, int] = {}
+    for lens in cfg.lenses:
+        if hasattr(lens, "n"):
+            out[lens.name] = lens.n[run_index]
+    return out
+
+
+def _worker_for(
```
f-wright: I don't understand the name here?

load-testing/app_benchmark/cli.py
```diff
@@ -0,0 +1,395 @@
...
+            )
+            p = mp.Process(target=target, kwargs=kwargs, name=f"{lens.name}_cam{cam_idx}")
+            p.start()
+            procs.append(p)
+            worker_number += 1
+    return procs
+
+
+def _emit_ramp_marker(log_file: Path, total_clients: int) -> None:
+    """Write one fake `RAMP <N> ts=<epoch>` line to the log so the
```
f-wright: Why would we write a fake RAMP marker instead of just writing a log indicating the main benchmark started? i think that could be confusing if anyone was looking at the logs to debug something

load-testing/app_benchmark/configs/example.yaml
honeytung: For me the `target_fps` set in the global level is for limiting all lenses to use a fixed fps so we don't need to manually add that for each lenses.

That option is overridable for each individual lenses if we need a lens that needs higher fps than others.

load-testing/app_benchmark/configs/example.yaml
honeytung: If we leave that option unset, it will use the global `target_fps` for this lens.

load-testing/app_benchmark/README.md
honeytung: Hmm maybe, the script uses the benchmark name in the yaml to generate the detector prefixes for each lens. So we could have the script just use the same matching names. We can expose an option to not do detector cleanup on the cloud. 

One thing we need to be aware is that we need to make sure the existing detector has the exact same `mlpipeline` that we specified in the yaml file so we don't think it is benchmark this pipeline but it actually isn't.

I planned for v2 of this script is to allow user to specify the detector ID directly so that will not get deleted and we can quickly run the benchmark.

load-testing/app_benchmark/README.md
honeytung: Makes sense, I have updated the README.

load-testing/app_benchmark/README.md
honeytung: This is what I observed from TDCJ simulations so I assume it applies to other pipelines as well. But I can check with the science team on this. Removing that line just in case.

load-testing/app_benchmark/configs/example.yaml
honeytung: We can make the default be `http://localhost:30101`, but I think it is good to leave it here as an example to point to different endpoints, since we need to specify that when we deploy this to test on a balena machine.

load-testing/app_benchmark/__init__.py
honeytung: I think python need this to be able to import files inside this folder.

load-testing/app_benchmark/cleanup_edge.py
honeytung: Good point, made `--list` be the default.

load-testing/app_benchmark/cli.py
honeytung: Yea, it was originally added because `parse_load_test_logs.py` expects that, but I let claude tune the `report.py` so we are not using that anymore. Removed.

load-testing/app_benchmark/README.md
honeytung: Removed this instructions, it should run without GPU but some metrics will not work but this is fine.

load-testing/app_benchmark/README.md
honeytung: This is recording a single inference request (i.e. there may be multiple request per frame for some lenses). Updated the docs to better reflect that.

load-testing/app_benchmark/cli.py
honeytung: Modified this to use a better name, this function is generating args for different lens modes and overwrite global args if necessarily.

load-testing/app_benchmark/README.md
honeytung: The `n` value controls the inference request on the subsequent binary detector so it always has fixed `n` amount of inference requests to that detector.

One thing to note is that we did not use the actual crops generated from the bbox detector, we use a fixed image for binary inference that got generated at work startup and perform `n` requests instead. I made that decision to minimize script load so we can have a better picture of the actual edge-endpoint performance, rather than depend on how fast the script can run (especially on the RPi).

Happy to hear what you think tho.

load-testing/app_benchmark/README.md
honeytung: Yea, I think that makes sense, it is not used for the subsequent inference request, but it may gives a more accurate inference time?

load-testing/app_benchmark/configs/example.yaml
f-wright: So it is a limit per lens, not per detector? I think my concern is when there are multiple detectors in a single lens, if both are limited to a particular target FPS the lens FPS will be below that

.gitignore
```diff
@@ -153,6 +153,8 @@ docker-compose.yml
 load-testing/logs/
 load-testing/plots/
 load-testing/load_tests/
+load-testing/benchmark-results/
```
f-wright: We should only need one of these, right?

load-testing/app_benchmark/README.md
f-wright: Yes, we definitely would need to verify the ML pipeline is correct. If you'd rather put this into a follow up PR, that's fine by me.

load-testing/app_benchmark/configs/example.yaml
honeytung: This is a limit per lens not per detector, so we can set what is the intended FPS for each lens. I think this is what we are trying to see: `Can a lens runs at some fps?`

load-testing/app_benchmark/config.py
```diff
@@ -0,0 +1,322 @@
...
+            warning and proceed, since `edge.set_config` merges cleanly
+            and restores the pre-run state at cleanup. Set True for CI
+            or any context where contamination is unacceptable.
+        set_config_timeout_seconds: How long to wait for `edge.set_config`
+            to finish (cold edges with many models need more).
+    """
+    name: str = Field(pattern=r"^[a-zA-Z0-9_-]+$", max_length=64)
+    output_dir: str = "./benchmark_results/{name}-{ts}/"
+    edge_endpoint_url: str
+    cloud_endpoint: str = "https://api.groundlight.ai/"
```
f-wright, line 50: Is this not configured by helm?

load-testing/app_benchmark/config.py
```diff
@@ -0,0 +1,322 @@
...
+    def _derive_prefix(self) -> "RunConfig":
+        if self.detector_name_prefix is None:
+            # Slug the name into the prefix shape: lowercase, dashes→underscores,
+            # strip anything outside [a-z0-9_], truncate to 16 chars, ensure
+            # it starts with a letter (fall back to "bench" if not).
+            slug = "".join(
+                c for c in self.name.lower().replace("-", "_")
+                if c.isalnum() or c == "_"
+            )[:16]
+            if not slug or not slug[0].isalpha():
```
f-wright: Should we error here instead or validate the prefix when the user provides it instead?

load-testing/app_benchmark/detectors.py
```diff
@@ -0,0 +1,301 @@
...
+        """Thin wrapper over groundlight_helpers.provision_detector that
+        tracks every created detector in `_all_created` for delete_all().
+
+        Args:
+            prefix: Detector name prefix from `_name_prefix(...)`.
+            mode: One of "BINARY", "BOUNDING_BOX", "COUNT", "MULTI_CLASS".
+            image_size: (width, height) used for the priming images.
+            pipeline: Optional pipeline config name; None uses the cloud
+                default for the mode.
+            n: Mode-specific knob (max_num_bboxes for BOUNDING_BOX,
```
f-wright, line 249: How does n interact with MULTI_CLASS? Can you specify the num classes like that or do you need to provide a list of class names elsewhere?

load-testing/app_benchmark/lenses.py
```diff
@@ -0,0 +1,289 @@
...
+
+    Args:
+        worker_number: Global worker index, recorded on every log line.
+        camera: This worker's per-lens camera index.
+        lens_name: Logged on every event; matched against config in the
+            report.
+        bbox_detector_id: Cloud detector ID for the upstream bbox stage.
+        binary_detector_id: Cloud detector ID for the downstream binary
+            stage.
+        n: Both the upper bound on object count in the bbox image AND
```
f-wright: What do you think about making `n` the exact number of objects in the bbox image? Would make it easier to control, and I don't think the max count should impact latency, only the actual number of objects (though maybe we should verify that)

load-testing/app_benchmark/report.py
```diff
@@ -0,0 +1,801 @@
...
+import json
+from collections import defaultdict
+from pathlib import Path
+from typing import Any
+
+import matplotlib.pyplot as plt
+from parse_load_test_logs import _percentile as _interp_percentile
+
+# Achieved FPS counts as "hit" target if it's within this fraction.
+_FPS_HIT_TOLERANCE = 0.95
```
f-wright, line 27: Why do we need a tolerance here? I would think we'd want to hit everything for it to count

load-testing/app_benchmark/detectors.py
```diff
@@ -0,0 +1,301 @@
...
+            elif isinstance(lens, BboxToBinaryLens):
+                assert max_n is not None
+                bbox_det = self._provision(
+                    prefix=_name_prefix(name_prefix, run_name, lens.name, "bbox"),
+                    mode="BOUNDING_BOX", image_size=image_size,
+                    pipeline=lens.bbox_pipeline, n=max_n,
+                )
+                bin_det = self._provision(
+                    prefix=_name_prefix(name_prefix, run_name, lens.name, "binary"),
+                    mode="BINARY", image_size=image_size,
```
f-wright: Looks like this is the bbox image size, not the binary downstream size, so the priming images would be different dimensions than the actually submitted ones during the benchmark

load-testing/app_benchmark/config.py
```diff
@@ -0,0 +1,322 @@
...
+        monitoring: SystemMonitor sampling settings.
+    """
+    model_config = ConfigDict(populate_by_name=True)
+    run: RunConfig
+    globals_: GlobalConfig = Field(alias="global")
+    lenses: list[LensSpec] = Field(min_length=1)
+    monitoring: MonitoringConfig = Field(default_factory=MonitoringConfig)
+
+
+def validate_config(cfg: BenchmarkConfig) -> None:
```
f-wright, line 244: There isn't any test coverage on these new additions, it seems like a couple simple ones on functions like this might be worth adding

load-testing/app_benchmark/report.py
```diff
@@ -0,0 +1,801 @@
...
+            f"> ⚠ **{len(failures)} worker process(es) exited with non-zero status**: "
+            + ", ".join(f"`{f['name']}` (exit={f['exitcode']})" for f in failures)
+        )
+    lines.append("")
+    lines.append("| Lens | Camera | Frames | Errors | FPS | Target | Hit | p50 (s) | p95 (s) |")
+    lines.append("|---|---:|---:|---:|---:|---:|:---:|---:|---:|")
+    for cam in s.get("cameras", []):
+        if cam.get("no_events"):
+            lines.append(
+                f"| {cam['lens_name']} | {cam['camera']} | 0 | ? | 0.0 | "
```
f-wright, line 552: Claude says this will be dropped in most renderings since it's a 10th entry in a table with a 9 column header. I haven't checked that, but maybe worth looking at if you haven't tested this part already?

load-testing/app_benchmark/lenses.py
```diff
@@ -0,0 +1,289 @@
...
+        "worker_number": worker_number,
+        "request_number": request_number,
+        "latency": round(end - request_start, 4),
+        "success": success,
+    }
+    if stage is not None:
+        record["stage"] = stage
+    if error is not None:
+        record["error"] = error
+    with open(log_file, "a", encoding="utf-8") as f:
```
f-wright: Is there a latency impact from opening and writing to this on every request? Claude recommended a per worker log file instead which would also prevent errors if multiple requests try to write at the same time here. I believe that's already how we do the metrics writes if you want to look there for a model

load-testing/app_benchmark/README.md
```diff
@@ -0,0 +1,191 @@
...
+```bash
+cd load-testing
+uv sync
+```
+
+## Run
+
+```bash
+cd load-testing
+GROUNDLIGHT_API_TOKEN=... python -m app_benchmark app_benchmark/configs/example.yaml
```
f-wright: Should this be `uv run python ...`? That's what worked for me.

Also, passing `GROUNDLIGHT_API_TOKEN` won't be necessary if the edge is already set up with one, right?

f-wright (approved):
Some more small comments, but in general I think this looks pretty good and we can get it in and start getting results!

A suggestion for another plot I think would be useful is a plot of achieved FPS / n value for a summary view. Really I'd like achieved FPS / num streams, but I think we decided rolling out over different numbers of streams was going to be added in the next version of this script.

If @timmarkhuff is available to review this PR as well, I think that'd be worthwhile since he's most familiar with the benchmarking pre-existing on the edge.

load-testing/app_benchmark/README.md
f-wright: Yeah, some (but not all) object detection models will have varying inference time based on the number of objects.

load-testing/app_benchmark/README.md
f-wright: That sounds good to me.

load-testing/app_benchmark/__init__.py
f-wright: Oh yeah sorry! I think they're just empty elsewhere in the edge-endpoint, but you're right that the file itself needs to remain

load-testing/app_benchmark/config.py
honeytung: The script ran its own SDK calls to `cloud_endpoint` when training a new detector with the specified ML pipeline. The next version we will add the ability to specify detector ID so we don't need to perform retraining.

.gitignore
honeytung: We need the folder `benchmark_results` because is where all the runs are located. I think the other three are the original `load_testing` script.

load-testing/app_benchmark/detectors.py
honeytung: Right now it only supports `single_bbox`, `bbox_to_binary`, and `single_binary`. The vision team said `bbox_to_multiclass` could be an additional mode that they are using. I can include that feature in V2.

load-testing/app_benchmark/lenses.py
honeytung: Make sense! I will add that to this PR.

load-testing/app_benchmark/report.py
honeytung: This is for the summary page to make sure the final average FPS is within that range. The decision to include this is because we used a simple `time.sleep(1/target_fps - processing_time)` to wait between each frame if the ML finishes early. From my testing it seems like sometimes it is not accurate enough so if `target_fps` is set to 5 it will be mostly around 4.9 (it will never reach exactly 5.0 or above). This adds a "tolerance" so we can account that.

load-testing/app_benchmark/README.md
honeytung: We still need to pass in `GROUNDLIGHT_API_TOKEN` as we are using the client SDK to process requests. The script is not running in helm.

load-testing/app_benchmark/lenses.py
```diff
@@ -0,0 +1,289 @@
...
+    while time.time() < deadline:
+        frame_start = time.time()
+        bbox_image, _, _ = imgh.generate_random_objects_image(w, h, max_count=n)
+        _submit_and_log(
+            gl, bbox_detector_id, bbox_image,
+            log_file=log_file, lens_name=lens_name, camera=camera,
+            worker_number=worker_number, request_number=request_number, stage="bbox",
+        )
+        request_number += 1
+        for _ in range(n):
```
f-wright: Do you know if the lenses run through the vision engine submit the downstream classification images in series like this or in parallel? 

We should probably match what they're doing for one benchmark, and if they're not submitting the binary requests in parallel benchmark that separately.

load-testing/app_benchmark/report.py
honeytung: Confirmed and fixed, thanks for catching that!

load-testing/app_benchmark/lenses.py
honeytung: I am not sure, I can check with them on this. I know they limit the number of classifications per frame.

.gitignore
f-wright: Sorry, I meant one of `load-testing/benchmark-results/` and `load-testing/benchmark_results/`

load-testing/app_benchmark/config.py
f-wright: Can we get the cloud endpoint url from the helm configuration though? If an edge endpoint is configured with an integ account, benchmarking would fail with this setup.

honeytung: Currently there is no way since edge-endpoint does not expose `upstreamEndpoint` through SDK nor status endpoints. I think that could be useful so I can create a ticket to expose that.

load-testing/app_benchmark/cleanup_edge.py
```diff
@@ -0,0 +1,86 @@
...
+
+    gl = ExperimentalApi(endpoint=args.edge_endpoint)
+    try:
+        current = gl.edge.get_config()
+    except Exception as exc:
+        logger.error("could not fetch current edge config: %s", exc)
+        return 2
+
+    det_entries = getattr(current, "detectors", None) or []
+    det_ids = [d.detector_id for d in det_entries if hasattr(d, "detector_id")]
```
timmarkhuff: Why are we using hasattr here? Seems like an unnecessary contingency. 

Generally I try to avoid hasattr at all, and here it doesn't seem to make sense.

load-testing/app_benchmark/lenses.py
```diff
@@ -0,0 +1,295 @@
...
+    n: int,
+    edge_url: str,
+    image_size: tuple[int, int],
+    target_fps: float,
+    duration_seconds: float,
+    log_file: str,
+) -> None:
+    """multiprocessing.Process target for a `single_bbox` lens worker.
+
+    Generates a fresh image containing up to `n` random objects per
```
timmarkhuff: This docstring is now stale. It's no longer a random image, but instead uses `generate_fixed_objects_image`

timmarkhuff (approved):
LGTM

load-testing/app_benchmark/lenses.py
honeytung: Good catch! Updated!

load-testing/app_benchmark/cleanup_edge.py
honeytung: Updated this and also `host_check.py` so it uses `d.detector_id instead`.

PR #411, opened 2026-05-19
Configurable timeouts for slow-network resilience (rollout 30m + startup 90m)

honeytung:
## Summary

Two related changes that make edge-endpoint adapt gracefully to slow-network environments. Both are pure-additive Helm values with sensible defaults — no behavior change required at the call site.

- **`modelUpdater.rolloutReadyTimeoutSeconds`** (default **1800s / 30 min**) — controls the model-updater's poll for an inference rollout to become Ready. Read from the `ROLLOUT_READY_TIMEOUT_S` env var in `app/model_updater/update_models.py`. Replaces a hardcoded 10-minute timeout that was causing memory cascades on slow networks.
- **`inferenceDeployment.startupProbe.failureThreshold`** (default **540 / 90 min**) — controls the kubelet startup-probe ceiling on each inference pod. Templated into `files/inference-deployment-template.yaml`. Replaces a hardcoded 270 (45 min) that wasn't enough for large multiclass backbones loading via mount-s3 under bandwidth contention.

## Why

Investigated a slow-network failure on g4 (2 MB/s ingress throttle, 9 detectors). Found two independent ceilings that interact:

### Ceiling 1 — model-updater 10-min poll (cycle-2 cascade)

1. Cloud frequently republishes a model binary for a detector (e.g. a frequently-retrained multiclass detector — observed ~10–45 min republish cadence for the `leadboardType` SigLIP detector in the test).
2. Model-updater patches the inference Deployment with a fresh `restartedAt` annotation. With `maxSurge=1, maxUnavailable=0` (preserving zero-downtime inference), K8s creates a new pod alongside the old one.
3. On a slow link, the new pod takes 10–25 min to cold-load. The hardcoded 10-min poll fires while the new pod is still loading. The model-updater logs `TimeoutError`, walks away, returns to the outer for-loop.
4. On the next outer-loop pass (~60s later), the cloud has often republished yet another newer binary. The updater patches again → *another* new pod alongside the still-loading previous one → memory doubles, then triples.
5. Node tips into `MemoryPressure: True` → kubelet evicts → ReplicaSet creates pods that can't schedule due to the memory-pressure taint → terminal-state pods accumulate → eventually the `edge-endpoint` control pod itself is evicted.

The fix is to set the poll timeout comfortably above realistic cold-load time so the for-loop self-throttles: each detector's rollout completes before the next starts.

### Ceiling 2 — inference pod startupProbe 45-min ceiling

The inference container reads pretrained backbone weights from `/opt/models/pretrained-weights/` (a mount-s3 FUSE filesystem, `HF_HUB_OFFLINE=1`). With multiple pods cold-loading concurrently, per-pod effective bandwidth shrinks; a SigLIP-V2 multiclass detector observed taking exactly 45 min in this state before kubelet killed it for failing the startup probe (270 × 10s = 45 min). The new container then restarted and resumed the same load, hitting the same ceiling again. RS generation count stayed at 1 (no cascade — ceiling-1 fix worked), but the pod itself can never make progress.

Bumping the failure threshold to 540 (90 min) handles the realistic mix on the test machine. Truly slow links (1 MB/s, cellular, satellite) can `--set` higher.

This is the trade-off the team accepted when switching from `sync-pinamod` (pre-download everything) to `mount-s3` (lazy-load on demand). mount-s3 wins on the common case (small detector, fast network) and storage cost, but shifts the network risk from deploy time to pod cold-start time, which lands on the inference pod's startup window.

## Empirical results on g4 (2 MB/s ingress throttle, 9 detectors)

| | Unpatched (run @ ~4h) | Patched (this PR) |
|---|---|---|
| ReplicaSet generations on det_38C primary | **13** | **1** |
| ReplicaSet generations on det_38C oodd | **11** | **1** |
| MemoryPressure | **True** (sustained) | **False** |
| Terminal-state pod husks | 41+ Error / Evicted / ContainerStatusUnknown | **0** |
| edge-endpoint control pod evictions | 2 (running on gen-3 replica) | **0** |
| `TimeoutError: ... not ready within time limit` log lines | many (every ~10 min) | **0** |

The remaining issue observed on g4 is a single pod (det_38C primary, SigLIP-V2) wedging at the 45-min startupProbe ceiling — exactly what the second change in this PR addresses.

## Choice of defaults

**30 min for model-updater rollout timeout:** covers realistic worst-case load times on the slowest links we want to support out of the box, and stays well below the inference startupProbe ceiling so the updater never sees a kubelet kill before the pod can become Ready.

**90 min for inference startupProbe:** covers SigLIP/DINOv2-class backbones loading via mount-s3 when bandwidth is shared across ~10–20 concurrent pods. Single-pod cold-load is much faster (~12 min at 2 MB/s); the high default accommodates the bandwidth-contention case.

The two values are coupled — model-updater timeout should always be < startupProbe ceiling, so kubelet doesn't kill a pod the updater is still waiting on. 30 min < 90 min ✓ Operators bumping one should bump the other proportionally.

## Test plan

- [x] Reproduced the cascade on g4 with the unpatched 0.7.1 chart at 2 MB/s ingress (13 RS generations for det_38C in ~4h, MemoryPressure=True, 41+ stale terminal pods).
- [x] Built a `:dev` edge-endpoint image with the patched code via `deploy/bin/build-local-edge-endpoint-image.sh`.
- [x] Confirmed `ROLLOUT_READY_TIMEOUT_S=1800` env var is present on the `inference-model-updater` container after `helm install ... --set edgeEndpointTag=dev` from the patched local chart.
- [x] Verified cycle 1 completes cleanly under the throttle: all 18 inference deployments created, no memory pressure, all RS at gen=1.
- [x] Verified `helm template ... --set inferenceDeployment.startupProbe.failureThreshold=720 --set modelUpdater.rolloutReadyTimeoutSeconds=3600` renders both overrides as expected.
- [x] `helm lint` passes on the chart.
- [ ] Apply chart-only update on g4 (`helm upgrade --reuse-values --set inferenceDeployment.startupProbe.failureThreshold=540`) and verify det_38C primary actually becomes Ready under the throttle.
- [ ] Long-run observation: confirm no ReplicaSet generation pile-up over multiple hours of cloud republish events.


app/model_updater/update_models.py
```diff
@@ -100,7 +105,7 @@ def _check_new_models_and_inference_deployments(
             and not deployment_manager.is_inference_deployment_rollout_complete(deployment_name=oodd_deployment_name)
         ):
             time.sleep(5)
-            if time.time() - poll_start > TEN_MINUTES:
+            if time.time() - poll_start > ROLLOUT_READY_TIMEOUT_S:
                 raise TimeoutError(f"Inference deployment(s) ({deployment_names}) are not ready within time limit")
```
timmarkhuff: Should we include the actual time limit here?

deploy/helm/groundlight-edge-endpoint/values.yaml
```diff
@@ -85,6 +85,29 @@ ecrRegistry: "767397850842.dkr.ecr.us-west-2.amazonaws.com"
...
+# inference-model-updater behavior
+modelUpdater:
+  # Seconds the model-updater waits for a new inference pod rollout to become Ready
+  # before logging a TimeoutError and moving on. Must comfortably exceed realistic
+  # cold-load time on the slowest link the device must support, so that the for-loop
+  # self-throttles back-to-back rollouts of the same detector (preventing memory
+  # doubling when cloud frequently republishes a model on a slow network).
+  # Stays below the inference startupProbe ceiling so the pod is never killed by
+  # kubelet before the updater sees it succeed.
+  rolloutReadyTimeoutSeconds: 1800  # 30 min
```
timmarkhuff, line 97: It seems this PR introduces the possibility for the user to provide invalid values. For example, `modelUpdater.rolloutReadyTimeoutSeconds` must be lower than `inferenceDeployment.startupProbe.failureThreshold` * 10. If it's not, I believe pods can never start up. Is there a way we could validate user input and fail loudly if the provide values don't make sense?

Claude recommends adding some code in in `_helpers.tpl` like:
```
{{/*
Validate that modelUpdater.rolloutReadyTimeoutSeconds is less than the
inference startupProbe ceiling (failureThreshold × 10s). If the rollout
timeout exceeds the probe ceiling, kubelet can kill an inference pod while
the model-updater is still waiting for it to become Ready.
*/}}
{{- define "groundlight-edge-endpoint.validateTimeouts" -}}
{{- $rollout := int .Values.modelUpdater.rolloutReadyTimeoutSeconds }}
{{- $ceiling := mul (int .Values.inferenceDeployment.startupProbe.failureThreshold) 10 }}
{{- if ge $rollout $ceiling }}
{{- fail (printf "modelUpdater.rolloutReadyTimeoutSeconds (%ds) must be less than inferenceDeployment.startupProbe.failureThreshold * 10s (%ds). Raise failureThreshold proportionally when increasing rolloutReadyTimeoutSeconds." $rollout $ceiling) }}
{{- end }}
{{- end }}
```

timmarkhuff (approved):
LGTM

I left a few comments.

deploy/helm/groundlight-edge-endpoint/values.yaml
honeytung: Makes sense! I updated `_helpers.tpl` with a new function `validate.timeouts` that checks the value.

PR #412, opened 2026-05-19
Add inference request and edge endpoint pod metrics to profiling dashboard

f-wright:
Surfaces wall time spent inside the edge-endpoint pod itself (request duration minus the union of inference-call intervals) as a derived row in the latency table, box plot, and histogram dropdown.

Also creates an inference request metric which excludes health checks and other short requests from the aggregation stats.


honeytung (approved):
LGTM!

PR #418, opened 2026-05-26
Add script to compare a pipeline's performance on edge vs on cloud

timmarkhuff:
Adds a script to run inference on Edge and Cloud using the same pipeline config in both places. Compares the resulting confidence and labels. In theory, results should be the same, so any large discrepancies would point to a issue with how we are performing inference, either on edge on on cloud. 

Edge inference is performed against the provided detector, running on edge configured with `NO_CLOUD`, so no data will be submitted to this detector. Cloud inference is performed against a new detector, primed from the provided detector; data will be submitted to this detector, but it doesn't matter, because it is a throw-away detector, created just for this test.


load-testing/compare_edge_cloud_inference.py
```diff
@@ -0,0 +1,507 @@
...
+    gl_cloud = ExperimentalApi(endpoint=glh.CLOUD_ENDPOINT_PROD)
+
+    pipeline = _select_pipeline(gl_cloud, detector_id)
+    pipeline_id = pipeline["id"]
+    mlb_key = pipeline["model_binary_id"]
+
+    source_detector = gl_cloud.get_detector(detector_id)
+    if source_detector.mode != "MULTI_CLASS":
+        print(
+            f"Error: source detector {detector_id} has mode {source_detector.mode!r}; "
```
brandon-wada, line 447: Is there a particular reason to only test multi-class?

load-testing/compare_edge_cloud_inference.py
```diff
@@ -0,0 +1,507 @@
...
+            "\n"
+            "Fetches recent image queries from the source detector, re-runs each image on\n"
+            "the edge endpoint (NO_CLOUD -- no cloud escalation) and on a fresh primed cloud\n"
+            "detector seeded from the detector's edge pipeline MLB, then prints a comparison\n"
+            "table of labels and confidence scores."
+        ),
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog=(
+            "environment variables:\n"
+            "  GROUNDLIGHT_API_TOKEN   Groundlight API token (required)\n"
```
brandon-wada, line 399: What permissions does the token need? Just to belong to the same account as the source detector?

brandon-wada (approved):

load-testing/compare_edge_cloud_inference.py
timmarkhuff: Yeah, I think due to the way we do permissions it has to be that way. I was hoping I could use an Admin account, but that didn't work.

load-testing/compare_edge_cloud_inference.py
timmarkhuff: Simply because that was the type of detector I was interested in at the time. Adding support for more detectors types would require more code (but not much more). YAGNI

If someone wants more types in the future, they can pretty easily extend this script.

PR #420, opened 2026-05-27
Lens Benchmarking v2: Support existing detector IDs, parallel training, and cameras & lens ramp

honeytung:
## Summary

v2 of the `app_benchmark` lens-sweep harness, building on the v1 base merged in #406. Five user-facing capabilities + one schema rename + a plot redesign:

1. **Existing detector IDs** (`bbox_detector_id`, `binary_detector_id`) — opt into pre-existing cloud detectors per stage. The harness fetches the detector by ID, verifies its actual pipeline matches the YAML's `pipeline:` field (fast-fails before any edge-config push if they differ), and preserves it at cleanup. Mix freely: one chained lens can use an existing bbox + freshly-trained binary, or vice versa.
2. **`run.preserve_detectors: bool`** — skip cloud-side deletion of detectors the harness created. Re-runs of the same config hit `get_or_create_detector` and reuse them; if the edge pipeline is already trained, priming is skipped entirely → provisioning becomes near-instant. Pipeline mismatches are still caught on the next run, so changing `pipeline:` between runs fails loudly.
3. **Parallel detector provisioning** — `provision_all` now uses the two-phase pattern PR #373 introduced (`provision_detectors`): phase 1 serially creates+primes each detector with `wait_for_training=False`, phase 2 waits for all to finish training. Total wall-clock is dominated by the slowest single detector, not the sum.
4. **Three independent sweep dimensions** — any of `objects`, `cameras`, `copies` can be a scalar (fixed) or a list (ramp). All list-typed fields across the config must share the same length and are **zipped** element-wise (run `i` uses element `i` of every list). The three dimensions answer different questions:
   - **`objects`** — per-frame workload. How many objects appear in each synthetic frame (and, for chained lenses, how many downstream binary calls run per frame). *(renamed from `n`)*
   - **`cameras`** — **how many clients hit a single lens.** N independent worker processes all submitting to the **same** detector(s). Measures concurrency against one model.
   - **`copies`** — **how many distinct lenses (detectors) the server juggles.** N independently-trained detector copies of the lens, each its own cloud-side detector with the same pipeline + objects setting. Measures how the edge scales as the number of *distinct loaded models* grows.

   So `cameras: [1,2,4]` keeps one detector and throws 1→2→4 clients at it; `copies: [1,2,4]` stands up 1→2→4 separate detectors (each with its own worker). They compose — `cameras: 2` + `copies: 4` runs 8 workers across 4 distinct detectors. `copies > 1` is incompatible with `*_detector_id` (a single pre-existing ID can't back multiple independently-trained copies) and the schema rejects the combination.
5. **Plot redesign** — `summary.md` embeds a new mosaic image (`fps_all_lenses.png`, 2-column grid of per-lens overlays) and per-lens detail plots (`fps_{lens}.png`). Each per-lens plot overlays one colored line per **camera** (when the lens has no copies-ramp) or per **copy** (when it does) via the viridis colormap — the dimension you're sweeping gets the focus — with aggregated error rate on a secondary axis. Per-(lens, copy, camera) PNGs still live on disk for ad-hoc inspection.
6. **`n` → `objects` rename** — the per-lens scalar/ramp field that controls "objects placed per synthetic frame" (and downstream binary calls for chained lenses) is now called `objects`, parallel with `cameras` and `copies`. Existing configs need this one-word edit.

All capabilities compose: external detectors, preserved detectors, parallel training, all three sweep dimensions, and the new plot layout are independent and work together.

### Schema diff (user-visible)

```yaml
run:
  # NEW
  preserve_detectors: false   # keep created detectors at cleanup for cross-run reuse

lenses:
  - name: door_lens
    cameras: 1                # OR cameras: [1, 2, 4, 8] — ramp CLIENTS hitting one detector

  - name: object_lens
    type: single_bbox
    objects: [1, 3, 5, 7]     # RENAMED from `n` — per-frame object count
    cameras: 1
    copies: 1                 # NEW; OR copies: [1, 2, 4, 8] — ramp DISTINCT detectors
    # NEW (optional)
    bbox_detector_id: det_xyz  # use existing detector, skip create+train (incompatible with copies > 1)

  - name: chain_lens
    type: bbox_to_binary
    objects: [2, 4, 6, 8]      # RENAMED from `n`
    # NEW (optional, independent per stage)
    bbox_detector_id: det_aaa
    binary_detector_id: det_bbb
```

### cameras vs copies at a glance

| | `cameras` ramp | `copies` ramp |
|---|---|---|
| What scales | Client processes per lens | Distinct detectors per lens |
| Detectors created | 1 (shared by all cameras) | N (one per copy, independently trained) |
| Question answered | Concurrency against one model | Capacity as distinct loaded models grow |
| Plot lines colored by | camera index | copy index |
| Works with `*_detector_id` | yes | no (rejected at validation) |

## Sample plots

fps_all_lenses
system_utilization

## Test plan

- [x] Pull this branch on G4 (`ssh g4 && cd ~/edge-endpoint && git fetch && git checkout htung_axoncorp/lens-benchmarking-v2`).
- [x] Run `~/.local/bin/uv run app_benchmark app_benchmark/configs/example.yaml` against the local edge endpoint. Expect 4 runs (sweep length 4 from `objects: [1,3,5,7]` and `objects: [2,4,6,8]`).
- [x] Confirm `summary.md` contains:
  - [x] "Lens configuration" table with one row per stage, `Objects` + `Cameras` + `Copies` columns showing list values
  - [x] "Overview" table with one row per run
  - [x] "Cross-run FPS" section with `fps_all_lenses.png` mosaic embedded + per-lens detail plots
  - [x] "Run details" sections with per-worker (Lens / Copy / Camera) latency rows
- [x] Confirm `plots/` contains `fps_all_lenses.png`, `fps_{lens}.png` for each lens, and the per-camera detail files under `plots/per_camera/`.
- [x] Set `preserve_detectors: true`, re-run the same config — provisioning should complete in seconds (detectors picked up by name, priming skipped).
- [x] Edit one lens to add a `binary_detector_id:` pointing at a pre-existing detector; re-run and confirm the harness logs `fetching external detector det_xxx for ...`, the Lens configuration table shows `(external)`, and the detector is preserved at cleanup.
- [x] **cameras ramp**: add `cameras: [1, 2, 4, 8]` on one lens (same length as the `objects` lists); confirm the Overview table gains a `lens_cameras` column, per-lens FPS plots show new camera lines starting partway through later runs, and boundary labels read `Run i (objects=X, cams=Y)`. Exactly one detector is created for that lens.
- [x] **copies ramp**: add `copies: [1, 2, 4, 8]` on one lens (same length); confirm `max(copies)` distinct detectors are provisioned at startup (one per copy, `_copy{k}` in the name), the Overview table gains a `lens_copies` column, the per-lens plot colors lines by **copy** (legend `copy 0`, `copy 1`, …), and detail files use the `fps_{lens}_copy{k}_camera_{N}.png` shape. Confirm `copies > 1` + `*_detector_id` on the same lens fails validation.


load-testing/app_benchmark/configs/example.yaml
```diff
@@ -1,44 +1,66 @@
...
+  output_dir: "./benchmark_results/{name}-{ts}/"    # where artifacts are written; {name} and {ts} are substituted at runtime, defaults to "./benchmark_results/{name}-{ts}/"
+  detector_name_prefix: "sample_bench"              # short prefix (≤16 chars) that all created detectors carry, defaults to a slug of `name` (lowercased, dashes → underscores)
+  refuse_if_host_not_clean: false                   # check if the edge-endpoint is not clean (has detectors configured) and stop if so, defaults to false
+  preserve_detectors: false                         # keep created detectors on the cloud after cleanup so re-runs reuse them via get_or_create_detector (skips retraining if the edge pipeline is already trained); pipeline mismatches are still caught at provision time; external (*_detector_id) detectors are always preserved regardless of this flag, defaults to false
+  # set_config_timeout_seconds: 900                 # how long to wait for edge.set_config to finish (cold edges with many models need more), defaults to 900
+  # cloud_endpoint: "https://api.groundlight.ai/"   # Groundlight cloud URL used for detector CRUD, defaults to "https://api.groundlight.ai/"
```
honeytung, line 9: I added all the fields that you can modify for the lens benchmarking script (commented out) just to make it clear what other options you can change.

timmarkhuff (approved):
LGTM, just a few comments. 

I wonder about the way we generate the number of objects for BOUNDING_BOX detectors. Always generating the same number of objects might not map well to the way an actual application works. That said, it's not obvious to me how we could do it better. 

I like the summary markdown file that is generated. It helps tie it all together. One thing that seems to be missing is a record of which machine the benchmark was run on. I gave some thought to how this could be automatically detected and recorded when I worked on `measure_ram_and_vram_usage.py`, and determined that automatic detection is too difficult and not worth it. Instead, I opted for --device-name being a required arg for the script, so at least some record is retained, even if it isn't perfectly exact. Maybe we want to add something like that here so we don't lose track of which machine a benchmark was run on? 

I noticed that it is not possible to specify a different image size for downstream binary detectors; they are always 224x224. I'm not sure that this is a problem, just calling that out. Maybe that should be recorded in the summary, in case we find that it is relevant later.

Nit: I see there is a hash of the run name is included in detector names. What is the reason for this? This means that if the exact same detector configuration (mode, pipeline, image size, etc.) is used in two different "runs", it will result in two different detectors being created. Not the end of the world, but it seems unnecessary.

PR #427, opened 2026-06-10
Interruptible `inference-model-updater` loop 

timmarkhuff:
## Background

When the `inference-model-updater` sleeps between update cycles, it does not respond to
`ExperimentalApi.edge.set_config` calls. If `refresh_rate` is high, this means adding or
removing detectors can be delayed by the full sleep duration.

## Changes

1. **Move `get_active_detector_ids()` onto `DatabaseManager`** as a class method. The
   module-level function in `edge_config_manager.py` delegates to it; no callers break.

2. **Replace `time.sleep` with `_wait_for_next_cycle`** in the model-update loop. Polls
   every 2s and wakes early if `set_config` added/removed a detector or `refresh_rate`
   changed. The detector baseline is captured once per cycle (before the per-detector loop)
   and shared with both the mid-cycle check and the wait, so a config change detected
   mid-cycle also causes `_wait_for_next_cycle` to return immediately rather than sleeping.

3. **Early-exit mid-cycle** — the per-detector for-loop checks after each detector and
   breaks early if the config changed, restarting the outer loop immediately.

## Notes

Paired with groundlight/python-sdk#439, which enforces a 1-day cap on `refresh_rate`.


configs/edge-config.yaml
```diff
@@ -1,7 +1,7 @@
 # For configuring detectors on the edge endpoint. See CONFIGURING-DETECTORS.md for more information.
 
 global_config: # These settings affect the overall behavior of the edge endpoint.
-  refresh_rate: 60 # How often to attempt to fetch updated ML models (in seconds). Defaults to 60.
```
timmarkhuff, line 4: The default is defined by the python-sdk. Better not to include it in a comment here. The comment could become stale.

pyproject.toml
```diff
@@ -12,7 +12,7 @@ dependencies = [
     "SQLAlchemy==2.0.35",
     "cachetools>=5.3.1,<6.0.0",
     "fastapi>=0.115.0,<0.116.0",
-    "groundlight>=0.25.0,<0.26.0",
+    "groundlight>=0.29.0,<0.30.0",
```
timmarkhuff: We were a bit behind on our python-sdk version

honeytung (approved):
LGTM!

PR #436, opened 2026-06-23
Serve status page edge-config under /status prefix

timmarkhuff:
## Problem

In the GEP hub dashboard, the Edge Endpoint Status widget showed "Failed to load." in its Configuration section, with a `GET /edge-config 404 (Not Found)` in the browser console. The other widgets (metrics, resources) loaded fine.

The cause: the status page is one React app, but its fetches don't all go to the same place. `metrics.json` and `resources.json` live under `/status/` and are served by the status-monitor container, while `/edge-config` is served by the main edge-endpoint API container at the root path. GEP's reverse proxy only forwards and rewrites the `/status` prefix into the hub, so `/edge-config` never matches a route and 404s at the GEP layer. Standalone, edge-endpoint's own nginx fronts both containers, so it works there.

## Fix

Serve a read-only copy of the active config from the status-monitor at `/status/edge-config` and point the frontend at it, so it rides the same proxy path as the other status data. No GEP change is required.

- `status_web.py`: add `GET /status/edge-config` returning `EdgeConfigManager.active().to_payload()` (the status-monitor already mounts the active-config PVC).
- `App.jsx`: fetch `/status/edge-config` instead of `/edge-config`.
- `vite.config.js` and `dev/mock_server.py`: update the dev proxy and mock to the new path.

The main API's `/edge-config` GET/PUT is unchanged for SDK and tooling use.

### DIFF
--- a/app/status_monitor/dev/mock_server.py
+++ b/app/status_monitor/dev/mock_server.py
@@ -359,7 +359,7 @@ class MockHandler(BaseHTTPRequestHandler):
                 data = build_resources(state)
             elif self.path == "/status/metrics.json":
                 data = build_metrics(state)
-            elif self.path == "/edge-config":
+            elif self.path == "/status/edge-config":
                 data = build_edge_config(state)
             else:
                 self.send_error(404)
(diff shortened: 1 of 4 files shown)


honeytung (approved):
LGTM!

PR #438, opened 2026-06-24
fix: canonicalize mis-cased detector IDs on the edge

timmarkhuff:
## Summary

Detector IDs are case-sensitive KSUIDs, but the cloud resolves them case-insensitively.
A mis-cased ID still resolves to the right detector and routes to the right inference pod,
but the edge's case-sensitive subsystems (hourly activity folders, edge-config lookup)
track it as a separate detector. This splits edge metrics into two buckets for one
detector, which then collide on the cloud's case-insensitive unique index
`(log_hour, group_id, device_id, detector_id)`, causing `IntegrityError` duplicate-key
failures and silent data loss.

## Fix

Resolve detector metadata up front and rebind `detector_id` to the canonical
`detector_metadata.id` before anything keys off it. Every downstream consumer (metrics,
edge-config lookup, inference routing, escalation) then uses one canonical casing,
regardless of caller spelling or whether the detector has an edge inference pod. This
mirrors the cloud's lenient resolution, replacing the earlier 404-rejection approach.

## Test plan

- [x] `test_post_image_query_canonicalizes_miscased_detector_id`: an upper-cased ID
      returns 200 and all metrics are keyed on the canonical `detector.id`.
- [x] Deploy changes to an edge endpoint and test that mixed-case submissions no longer split metric buckets.

### DIFF
--- a/app/api/routes/image_queries.py
+++ b/app/api/routes/image_queries.py
@@ -115,9 +115,9 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
     # Ensure that detector_id has correct casing by pulling the detector ID out of detector_metadata
     # get_detector_metadata returns the correctly-cased, canonical detector ID
     detector_metadata = get_detector_metadata(detector_id=detector_id, gl=gl)  # NOTE: API call (once, then cached)
-    detector_id = detector_metadata.id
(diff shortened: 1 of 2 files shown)


honeytung (approved):
LGTM!

PR #439, opened 2026-06-26
Make status page detector links point to the correct dashboard per environment

timmarkhuff:
## Problem

Detector ID links on the "Groundlight Edge Endpoint Status" page hardcoded `https://dashboard.groundlight.ai`, so they only resolved correctly on legacy prod. On integ/dev/us1 they pointed at the wrong dashboard.

## Solution
This PR derives the Cloud Dashboard base URL at runtime from the Groundlight SDK client, which already knows which cloud this device is registered to. It swaps the leading `api.` host label for `dashboard.`, so links resolve correctly everywhere:

- `api.groundlight.ai` -> `dashboard.groundlight.ai`
- `api.integ.groundlight.ai` -> `dashboard.integ.groundlight.ai`
- `api.dev.groundlight.ai` -> `dashboard.dev.groundlight.ai`
- `api.groundlight.dev.axon.com` -> `dashboard.groundlight.dev.axon.com`
- `api.groundlight.us1.axon.com` -> `dashboard.groundlight.us1.axon.com`

## How it works

- `status_web.py` (the web layer that backs the status page) gains a `cloud_dashboard_url()` helper and a `GET /status/cloud-config` endpoint. The helper reads the endpoint from the EE-wide SDK client and applies the `api.` -> `dashboard.` swap. It always builds an `https` URL (every Groundlight cloud is served over https, including eksdev) and raises if the client endpoint has no host, rather than emitting a malformed URL. Keeping this in `status_web.py` (rather than the cloud-bound metrics payload) avoids polluting cloud telemetry with a UI string.
- The React app fetches `/status/cloud-config` once on mount and passes the base URL to `DetectorDetails`, which builds `${dashboardUrl}/reef/detectors/${id}`. It falls back to `https://dashboard.groundlight.ai` if the fetch fails.
- The detector links already opened in a new tab; added `rel="noopener noreferrer"` for safety.
- Dev tooling (mock server + Vite proxy) updated so the new endpoint works locally.

## Deployment fix (required for this to work)

The `status-monitor` container, which serves `/status/cloud-config`, was the only cloud-talking container in the Helm chart without `GROUNDLIGHT_ENDPOINT` injected. Without it, `Groundlight()` defaulted to prod, its connectivity check rejected the non-prod token, `/status/cloud-config` returned HTTP 500, and the page silently fell back to the hardcoded prod dashboard. The chart now injects `GROUNDLIGHT_ENDPOINT` (from `upstreamEndpoint`) into the status-monitor container, matching the other containers.

This same gap was also silently breaking hourly metric reporting from the status-monitor on non-prod deployments (it used the same client against prod with a non-prod token), so this fix repairs that too.

## Refactor

Consolidated three near-duplicate Groundlight client factories into a single cached, endpoint-wide `app/core/groundlight_client.py::groundlight_client()`. `status_web.py`, `metric_reporting.py`, and `manage_reader.py` now share it, removing the duplicated private `_groundlight_client` helpers (and the cross-module import of a private symbol). Test patch targets updated accordingly.

## Test plan

- [x] On a non-prod deployment (dev), confirm `/status/cloud-config` returns the matching `dashboard.*` host and detector links open the detector page on that dashboard in a new tab.
- [x] Confirm prod behavior is unchanged (`dashboard.groundlight.ai`).

### DIFF
--- a/app/core/groundlight_client.py
+++ b/app/core/groundlight_client.py
@@ -1,0 +1,15 @@
+from functools import lru_cache
+
+from groundlight import Groundlight
+
+
+@lru_cache(maxsize=1)
+def groundlight_client() -> Groundlight:
+    """Return a cached, endpoint-wide Groundlight client for talking to the cloud.
+
+    Authenticates with the endpoint's environment-provided API token and cloud endpoint
+    (GROUNDLIGHT_API_TOKEN / GROUNDLIGHT_ENDPOINT), so it always points at the cloud this
+    device is registered to.
+    """
+    # Don't specify an API token here - it will use the environment variable.
+    return Groundlight()  # NOTE this will wait the default 10 seconds when there's no connection.
--- a/app/escalation_queue/manage_reader.py
+++ b/app/escalation_queue/manage_reader.py
@@ -2,13 +2,13 @@ import json
 import logging
 import os
 import time
-from functools import lru_cache
 from pathlib import Path
 
 from fastapi import HTTPException, status
-from groundlight import Groundlight, GroundlightClientError, ImageQuery
+from groundlight import GroundlightClientError, ImageQuery
 from urllib3.exceptions import MaxRetryError, ReadTimeoutError
 
+from app.core.groundlight_client import groundlight_client
 from app.core.utils import safe_call_sdk
 from app.escalation_queue.failed_escalations import record_failed_escalation
 from app.escalation_queue.models import EscalationInfo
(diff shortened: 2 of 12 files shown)


app/core/groundlight_client.py
```diff
@@ -0,0 +1,15 @@
+from functools import lru_cache
+
+from groundlight import Groundlight
+
+
+@lru_cache(maxsize=1)
```
timmarkhuff, line 6: Factoring this out into a core service since we use it in several places

honeytung (approved):
LGTM!

PR #441, opened 2026-06-29
fix: bump base images to Debian 12 (bookworm) — main + bastion

CoreyEWood:
## Why

The Snyk Container findings on `edge-endpoint-snapshot` (Crit/High) trace almost entirely to **Debian 11 (bullseye)** base images, which are frozen — many CVEs show as *"No Fix Supported"* only because the patched packages live in newer Debian.

Two bullseye bases:
1. `Dockerfile:25` — `python:3.11-slim-bullseye` (main edge-endpoint image)
2. `deploy/balena-k3s/bastion/Dockerfile:15` — `debian:bullseye-slim` (bastion)

## What

Both bumped to Debian 12 (bookworm):
- `python:3.11-slim-bullseye` → `python:3.11-slim-bookworm`
- `debian:bullseye-slim` → `debian:bookworm-slim`

## Verification

Built both targets locally (linux/arm64) — apt resolves clean, deps install, Debian 12.14. Confirmed by a fresh Snyk rescan of the main-image bump alone:

| | Before | After main bump | After bastion (expected) |
|---|---|---|---|
| Total issues | 495 | 430 | ~348 |
| Critical | 9 | 5 | ~2 |
| High | 33 | 15 | ~15 |
| Container findings | 322 | 257 | ~175 |

The rescan showed Debian IDs flipping `SNYK-DEBIAN11-*` → `SNYK-DEBIAN12-*` on the main image; the bastion bump clears the remaining ~82 Debian-11 findings + 3 bastion Criticals.

## Breaking-change risk assessment

**Why we expect this to be safe:** the image tag keeps **Python 3.11** (`python:3.11-slim-bullseye` → `-bookworm`), so the Python runtime and all pinned deps are unchanged. Only the **OS layer moves (Debian 11 → 12)** — system libraries and apt package versions.

| OS change | bullseye → bookworm | Risk |
|---|---|---|
| **OpenSSL** | 1.1.1 → 3.0 | **Highest.** Stricter TLS — rejects legacy/weak ciphers, tiny RSA keys, some renegotiation. Fine for normal HTTPS; only breaks against peers using deprecated crypto. |
| glibc | 2.31 → 2.36 | Low. Backward-compatible; binaries built on bookworm run fine. |
| nginx | 1.18 → 1.22 | Low-med. Config largely compatible; a rare deprecated directive could fail to load. |
| sqlite | 3.34 → 3.40 | Very low. File format stable, backward-compatible. |
| bastion `python3` | 3.9 → 3.11 | Low. Only matters if a bastion script relied on 3.9 behavior. |

**Already de-risked:** both images **build** and **apt resolves every package** on bookworm — so the most common upgrade failure (removed/renamed packages) is ruled out.

**Relies on CI to confirm runtime behavior:** the local build proves the image assembles, not that the app serves. The PR's integration tests + a container smoke test (health endpoint, an inference call exercising TLS/OpenSSL 3.0, nginx serving the status page) are the real signal before merge.

## Residual (not fixed here)
- **CVE-2023-45853** (zlib) still flags on the bookworm main image — *No Fix Supported* on Debian 12. Verified false positive (MiniZip not compiled into Debian `zlib1g`); risk-accepted in #444 (stacked on this branch).
- CVE-2023-4863 (libwebp, canary deps) cleared by #440.


honeytung (approved):
LGTM!

PR #442, opened 2026-06-29
fix: cap inference-pod CPU threads and offload blocking event-loop calls

honeytung:
## Summary

- **Root cause**: inference pods had no CPU thread limits, so each pod spawned ~112 OMP threads. With N detector copies = 2N pods, total threads exceeded 1000 on a 112-core machine — classic congestion collapse. Throughput peaked at 88 req/s (3 copies) then fell while CPU cost per request grew 5×; GPUs stayed at 2–5%.
- **Deployment fix**: cap the inference pod's CPU threads via a configurable Helm value — `inferenceDeployment.cpuThreads` (default `4`) sets `OMP_NUM_THREADS`/`MKL_NUM_THREADS`/`OPENBLAS_NUM_THREADS`. Inference pods set **no CPU requests or limits (BestEffort)**, so the scheduler launches every detector regardless of core count; threads are bounded by `cpuThreads`, which also avoids CFS throttling.
- **App fix**: add `timeout=2` to the no-timeout `is_edge_inference_ready` health check; offload `get_detector_metadata`, `inference_is_available`, and `run_inference` from the async event loop via `run_in_threadpool`; make `get_app_state` `async def` to remove the pointless threadpool dispatch on every request.

## Changes

**`deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml`**
- Set `OMP_NUM_THREADS / MKL_NUM_THREADS / OPENBLAS_NUM_THREADS` from `inferenceDeployment.cpuThreads` (default `4`) — caps each pod's intra-op thread pool so N pods don't oversubscribe the cores. Tune per device.
- Set **no CPU requests or limits** on the inference pods (**BestEffort**), so the scheduler launches every detector regardless of core count (detectors are mostly idle; active pods stay bounded by `cpuThreads`, and the OS time-slices simultaneous bursts). No CFS throttling.

**`deploy/helm/groundlight-edge-endpoint/values.yaml`**
- Add `inferenceDeployment.cpuThreads` (default `4`) so the thread cap is tunable per device without editing the chart.

**`app/core/edge_inference.py`**
- Add `timeout=2` to `requests.get()` in `is_edge_inference_ready` — previously no timeout; a briefly-unreachable pod could block the event loop for up to 60s. `Timeout` is a subclass of `RequestException`, so the existing `except` already handles it.

**`app/api/routes/image_queries.py`**
- `await run_in_threadpool(get_detector_metadata, ...)` — was a sync network call on the event loop (cached after first call, but first call can take 2–3s).
- `await run_in_threadpool(inference_is_available, ...)` — was 2 sync `requests.get` health checks on the event loop, each previously with no timeout (p99=1023ms in profiling).
- `await run_in_threadpool(run_inference, ...)` — was blocking `f_primary.result() / f_oodd.result()` waits on the event loop.

**`app/core/app_state.py`**
- `get_app_state`: `def` → `async def` — the function is a pure attribute read; FastAPI was dispatching it to a threadpool on every single request unnecessarily.

## Test plan

- [x] Run `make test` — 342 unit tests pass, pre-existing PermissionError fixture failures are unrelated to these changes
- [x] On device: `kubectl exec -n edge <inference-pod> -- sh -c 'echo OMP=$OMP_NUM_THREADS; python -c "import torch; print(torch.get_num_threads())"'` — expect 4
- [x] On device: confirm the rendered inference template caps threads (`OMP_NUM_THREADS=4`) and sets no CPU `resources` block (BestEffort)
- [x] Re-run the lens benchmark sweep (`copies: [1,2,3,4,5]`, `target_fps: 0`) and confirm throughput no longer collapses past 3 copies, per-request CPU cost stays roughly flat, and GPU% rises

---

## Validated on hardware

Device `avc2-profiling` (2× Xeon Gold 5520+ / 112 threads, 4× RTX Pro 4000 Blackwell). Lens benchmark `bbox_to_binary`, `objects=5`, `target_fps=0` (saturate), copies 1→5:

| metric (at 5 copies / saturating) | pre | post |
|---|---|---|
| total throughput | 72 req/s | **245 req/s (3.4×)** |
| scaling vs detectors | plateaus then collapses (8.4→14.6→12.0 FPS) | scales linearly (9.9→40.8 FPS) |
| p95 latency | 281 ms | **74 ms (3.8× lower)** |
| CPU cost / request | 1.25 core-sec | **0.04 core-sec (32× less)** |
| node CPU | 80% | 8% |
| GPU utilization | 4.6% | 11.3% |

Root cause was PyTorch/OpenMP **thread oversubscription** in the inference pods (no thread cap → each pod spawned ~112 threads; N pods thrashed the cores). The thread cap + event-loop offload eliminate it; per-request CPU cost is now flat with load (the oversubscription signature is gone) and the box is no longer CPU-bound.


app/api/routes/image_queries.py
```diff
@@ -114,7 +115,9 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
 
     # Ensure that detector_id has correct casing by pulling the detector ID out of detector_metadata
     # get_detector_metadata returns the correctly-cased, canonical detector ID
-    detector_metadata = get_detector_metadata(detector_id=detector_id, gl=gl)  # NOTE: API call (once, then cached)
+    detector_metadata = await run_in_threadpool(
```
timmarkhuff, line 118: This seems unnecessary, but maybe it doesn't hurt. 

I ran a load test with this code and didn't see any issues.

app/api/routes/image_queries.py
```diff
@@ -173,11 +176,15 @@ async def post_image_query(  # noqa: PLR0913, PLR0915, PLR0912
         # If human review is required, we should skip edge inference completely
         logger.debug("Received human_review=ALWAYS. Skipping edge inference.")
         record_activity_for_metrics(detector_id, activity_type="escalations")
-    elif app_state.edge_inference_manager.inference_is_available(detector_id=detector_id):
+    elif await run_in_threadpool(app_state.edge_inference_manager.inference_is_available, detector_id=detector_id):
```
timmarkhuff, line 179: Eventually I would like to get rid of this readiness cache. I have a PR that should do the trick: https://github.com/groundlight/edge-endpoint/pull/419

But for now this seem fine.

deploy/helm/groundlight-edge-endpoint/files/inference-deployment-template.yaml
```diff
@@ -75,6 +75,17 @@ spec:
           value: "1"
         - name: EDGE_COMPILE_ENABLED
           value: "{{ .Values.edgeCompileEnabled }}"
+        # Cap PyTorch/OpenMP/MKL/OpenBLAS intra-op threads so that many inference pods
+        # don't oversubscribe the node's cores. See values.yaml inferenceDeployment.cpuThreads.
+        - name: OMP_NUM_THREADS
```
timmarkhuff, line 80: Good work!

timmarkhuff (approved):
LGTM

PR #444, opened 2026-06-30
chore(snyk): risk-accept CVE-2023-45853 (zlib MiniZip false positive)

CoreyEWood:
## What

Adds a `.snyk` policy ignoring **CVE-2023-45853** (Critical) on `zlib1g`, with a 1-year expiry.

> **Stacked on #441** (bookworm base-image bumps). Merge #441 first. This branch reflects the post-bump state — both main and bastion images on Debian 12, so a single ID covers both.

## Why this is the right call (not a code fix)

The vulnerability is an integer overflow in **MiniZip**, a contrib component in the zlib source tree. Debian does **not** compile MiniZip into the shipped `zlib1g` package — so the vulnerable code is **absent from our images**.

- Debian security tracker marks `zlib1g` as **not affected**; no fix will be issued.
- Hence Snyk's permanent *"No Fix Supported"* — bumping the base image cannot clear it.
- Ref: https://security-tracker.debian.org/tracker/CVE-2023-45853

Risk-accepted with rationale + expiry rather than left as a noisy unfixable Critical.

## Scope

Ignores `SNYK-DEBIAN12-ZLIB-6008963` — the Debian 12 ID for both the main `Dockerfile` and `deploy/balena-k3s/bastion/Dockerfile` once #441's bookworm bumps are in.

## Context

Final item retiring the container Criticals. After #440 (canary removal, merged), #441 (bookworm bumps), and this ignore → real remaining Criticals = **0**.


honeytung (approved):
LGTM!

PR #448, opened 2026-07-01
Pin edge self-signed TLS cert to the device instead of regenerating on every boot (GL-1699)

timmarkhuff:
## Summary

The Lens<->edge nginx TLS cert was written to an `emptyDir` and regenerated by the `generate-tls-cert` initContainer on **every pod start**. This meant the edge endpoint's identity changed on each restart, forcing clients to disable certificate verification (a MITM/CJI exposure per GL-1699).

This PR "pins" the self-signed cert to the device so it stays stable across restarts. Per the direction on the ticket, this is the minimal EA fix; CA chaining, hostname verification, Lens-side trust distribution, and mTLS/endpoint auth are intentionally deferred to follow-up work.

## Changes

- **Persist certs to device-scoped storage.** The `nginx-certs` volume changes from `emptyDir` to a `hostPath` at `/opt/groundlight/device/certs` (matching the existing device-identity `hostPath` pattern), in both the Helm deployment template and the legacy k3s manifest.
- **Make cert generation idempotent and self-healing.** `generate-tls-cert.sh` now reuses the existing cert only when both files exist, the cert is valid for at least 30 more days, **and** the private key matches the certificate. A missing, mismatched, or near-expiry pair is regenerated (loudly) rather than left for nginx to choke on later. `CERT_DIR` is now overridable to support testing.
- **Add regression tests** (`test/bin/test_generate_tls_cert.py`) covering cert stability across repeated generator runs and self-healing of a mismatched cert/key pair.

## Testing

- New unit tests pass in ~0.6s (`uv run pytest test/bin/`).
- Manual verification: capture the served cert fingerprint (`openssl s_client -connect <edge>:443 | openssl x509 -noout -fingerprint -sha256`), restart the `edge-endpoint` pod, and confirm the fingerprint is unchanged.
- added `test/bin/test_generate_tls_cert.py`. Both tests passed. `test_mismatched_key_is_replaced` in ~0.52s and `test_valid_cert_is_pinned_across_runs` in ~0.35s
## Notes

- Existing edge devices will generate a fresh cert once on their next restart after this rolls out (no cert exists on the hostPath yet); from then on it is pinned and stable.
- The cert is valid 365 days and regenerates within 30 days of expiry, so there is no long-term expiry cliff.

## Jira Ticket
https://taserintl.atlassian.net/browse/GL-1699

### DIFF
--- a/app/bin/generate-tls-cert.sh
+++ b/app/bin/generate-tls-cert.sh
@@ -2,8 +2,25 @@
 
 # This script generates a self-signed TLS certificate and key to be used
 # by nginx to enable HTTPS.
+#
+# The certificate is "pinned" to the device: it lives on persistent storage and
+# is only (re)generated when it is missing or within 30 days of expiry. Keeping
+# it stable across pod restarts gives the Lens<->edge link a consistent endpoint
+# identity instead of a brand-new cert on every boot.
 
 set -ex
+
+CERT_DIR=/etc/nginx/certs
+CERT_FILE="$CERT_DIR/certificate.crt"
+KEY_FILE="$CERT_DIR/private.key"
+
+# Reuse the existing cert if it is present and valid for at least 30 more days.
+if [ -f "$CERT_FILE" ] && [ -f "$KEY_FILE" ] && openssl x509 -checkend 2592000 -noout -in "$CERT_FILE"; then
+    echo "Existing TLS certificate is still valid; keeping the pinned cert."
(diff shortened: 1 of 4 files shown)


tomfaulhaber (commented):
Don't you need to force certificate regeneration periodically to make sure it happens even if the edge endpoint is running for a very long time?

tomfaulhaber (approved):

PR #450, opened 2026-07-07
feat: enforce 7-day retention policy on all escalation-queue data

CoreyEWood:
## Summary

EA security requirement: escalation-queue data must have a defined retention policy. This PR enforces a **hard 7-day retention on everything the escalation queue writes to disk** — images, pending escalation data, and failed-escalation records — deleted purely by age, **regardless of escalation status**.

- Add `app/escalation_queue/queue_retention.py`: `prune_expired_queue_data()` deletes any file older than `QUEUE_RETENTION_DAYS = 7` (by mtime) across the `images/`, `writing/`, `reading/`, and `failed/` directories. It also exposes a `python -m app.escalation_queue.queue_retention` entrypoint that runs a single sweep and exits.
- Run it from a dedicated **Kubernetes CronJob** (`deploy/helm/groundlight-edge-endpoint/templates/queue-retention.yaml`, `schedule "0 4,16 * * *"`, `concurrencyPolicy: Forbid`), launched via `app/bin/launch-queue-retention.sh`, mounting the same `escalation-queue-volume` hostPath as the reader.
  - **CronJob, not an in-process scheduler**: the schedule is wall-clock owned, so a restart or crash-loop of the edge-endpoint pod can't reset a timer and silently starve the retention guarantee. Keeping pruning out of `status_web.py` also keeps that container focused on observability.
  - Twice daily keeps the worst-case on-disk age within ~half a day of the window.
- `prune_failed_escalations()` keeps only its `MAX_RECORDS=200` count cap and tmp-file cleanup (still run on every failed write); its age logic is now owned by the unified sweep.
- `QUEUE_RETENTION_DAYS` lives in `constants.py` so the sweep and the reader share one source.

### Don't resurrect expired data as new failure records
Deleting a stale escalation's image can make the in-flight escalation attempt fail with a missing-image error. Left unhandled, that would write a **new** failed-escalation record (fresh timestamp) for data we're actively purging — extending that escalation's metadata lifetime past the window.

Fix: when an escalation fails **and** it is already past the retention window, drop it without recording (`manage_reader._escalation_is_expired`, age taken from the escalation's own `timestamp`). This is exact, not heuristic — the sweep never deletes anything younger than the window, so a missing image on an in-window escalation is a genuine failure and is still recorded. Malformed-line failures (no parseable age) are also still recorded; parse errors fail toward recording.

### Robustness / edge cases
- **Reader crash-loop guard.** The sweep deletes `reading/` files by mtime independently, so a stalled reader crossing the window could leave a data file deleted while its (newer) tracking file survives. `QueueReader` would then keep re-selecting the missing data file and raise `FileNotFoundError` out of its generator on every restart. `__iter__` now guards this: if the data file is gone, it discards the orphaned tracking file and moves on. (Regression test fails without the guard.)
- **`writing/` rename race.** The sweep runs in a separate process and can delete a `writing/` file (if >7 days old) between the reader listing the dir and renaming the chosen file into `reading/`, raising `FileNotFoundError` out of `_choose_new_file`. That is now caught and treated as "no file available" (return `None`, poll again); the writer already guards the same race at `queue_writer.py:84`. (Regression test added.)
- **Best-effort sweep.** Each retention dir is listed inside a `try/except` so a dir vanishing mid-sweep skips that dir rather than aborting the whole run; per-file errors were already caught.
- **Retryable-failure caveat.** The drop-without-recording decision only governs *permanent* failures; retryable failures loop in `consume_queued_escalation` and never surface, so an expired escalation stuck on a retryable error is neither dropped nor recorded until it fails permanently (documented at `_escalation_is_expired`).

### Design note
Deletion is intentionally **status-independent**: stale pending escalations older than 7 days are dropped along with orphaned images and failed records. This is a deliberate data-retention bound, not just orphan cleanup — an escalation sitting in the queue for a week is treated as expired data. Age is taken from file mtime for the sweep, and from the escalation's queued timestamp for the record-skip decision.

## Test plan

`test/escalation_queue/test_queue_retention.py` (ages keyed off `QUEUE_RETENTION_DAYS`, so they hold if the window changes):
- [x] `test_deletes_data_older_than_window_in_every_dir` — old files pruned in images / writing / reading / failed
- [x] `test_keeps_data_within_window` — files newer than the window preserved in every dir
- [x] `test_deletes_pending_escalation_regardless_of_status` — old pending escalation + its image removed even though never escalated
- [x] `test_ignores_subdirectories` — non-file entries skipped without error
- [x] `test_missing_dir_is_skipped` — a nonexistent retention dir is skipped, not fatal

`test/escalation_queue/test_escalation_queue.py`:
- [x] `test_expired_escalation_failure_is_not_recorded` — failure of a past-window escalation is dropped, not recorded
- [x] `test_recent_escalation_failure_is_recorded` — failure of an in-window escalation is still recorded
- [x] `test_reader_skips_orphaned_tracking_file` — reader discards an orphaned tracking file instead of crashing (fails without the guard)
- [x] `test_choose_new_file_handles_rename_race` — reader treats a file deleted between listing and rename as "no file available" instead of crashing
- [x] Existing `TestReadFromEscalationQueue` / `TestQueueReader` / `TestPruneFailedEscalations` still pass


app/status_monitor/status_web.py
```diff
@@ -56,6 +57,10 @@ async def startup_event():
     # seconds, to avoid every edge-endpoint report hitting the server at the exact same time.
     scheduler.add_job(reporter.report_metrics_to_cloud, "cron", hour="*", minute="3", jitter=120)
     scheduler.add_job(clear_old_activity_files, "interval", seconds=ONE_HOUR_IN_SECONDS)
+    # Enforce the queue data-retention window (images + escalation data + failed records). cron (not
+    # interval) so restarts don't reset the timer and starve the job; twice daily keeps the worst-case
+    # on-disk age within ~half a day of the window.
+    scheduler.add_job(prune_expired_queue_data, "cron", hour="4,16")
```
CoreyEWood: This seemed like a sensible place to put this to avoid having to create a whole new mechanism for the job. But do let me know if it seems bad to you.

test/escalation_queue/test_queue_retention.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def _write(path: Path, age_days: float) -> Path:
+    path.write_text("data")
+    ts = (datetime.now(timezone.utc) - timedelta(days=age_days)).timestamp()
+    os.utime(path, (ts, ts))
+    return path
+
+
+def test_deletes_data_older_than_window_in_every_dir(dirs):
```
timmarkhuff: Should we use type annotation in these function signatures?

app/status_monitor/status_web.py
timmarkhuff: At first glance, this does seem weird since it doesn't seem to relate to "status_web". This whole file/container is already a bit of a mixed bag. Claude helped me summarize everything that is going on here:

1. Metrics to cloud: collect_metrics_for_cloud, report_metrics_to_cloud.
2. Metrics-file housekeeping: clear_old_activity_files (GCs the IQ activity files that feed the metrics). This is metrics-adjacent, so it fits the theme.
3. Serving a status/observability web UI + JSON: /status React page, /status/metrics.json, /status/resources.json, /status/edge-config, /status/cloud-config, static assets.
4. Escalation-queue data retention: prune_expired_queue_data. (proposed in this PR)

The overall theme seems to be observability, so a sensible refactor (maybe not for this PR) might be to rename the container `status-monitor` to `observability` and the file `app/status_monitor/status_web.py` to `app/observability/server.py`.

But now I am feeling that adding pruning expired escalations _is_ pretty irrelevant here.

I think it might be better to just execute your pruning code with a k8s cronjob, rather than reusing the scheduling machinery in status_web.py.

app/escalation_queue/manage_reader.py
```diff
@@ -120,11 +142,17 @@ def read_from_escalation_queue(reader: QueueReader, request_cache: RequestCache)
             else:
                 logger.debug("Duplicate request ID received: %s. Skipping", escalation_info.request_id)
         except Exception as e:
-            logger.error("Escalation failed, moving on.", exc_info=True)
+            if escalation_info is not None and _escalation_is_expired(escalation_info):
+                # The escalation is already past the retention window, so its data is being purged
+                # regardless. Don't record the failure: the record isn't actionable (the escalation
+                # can't be retried) and would just re-add expired data with a fresh retention clock.
+                logger.warning("Escalation failed but is past the retention window; dropping without recording.")
```
f-wright, line 149: It seems like this should also be an error level log, if failing and a later retry is at level error?

app/escalation_queue/queue_reader.py
timmarkhuff: Claude pointed out this "medium" severity issue:
_New race: the sweep can now delete files from writing/ between the sorted() above and this rename, which would raise FileNotFoundError and crash the reader process. Low probability (file has to be >7 days old) and self-healing on restart, but worth a try/except FileNotFoundError + continue here to be safe. The writer already does this at queue_writer.py:84._

app/escalation_queue/manage_reader.py
```diff
@@ -120,11 +142,17 @@ def read_from_escalation_queue(reader: QueueReader, request_cache: RequestCache)
...
+            else:
+                logger.error("Escalation failed, moving on.", exc_info=True)
 
-            # Save the failed escalation details to disk. Once we integrate with Splunk, we could consider
-            # removing this and just relying on the error log instead.
-            record_failed_escalation(escalation, e)
+                # Save the failed escalation details to disk. Once we integrate with Splunk, we could consider
```
f-wright, line 153: Is there a jira ticket for this? Seems like it could get lost in the comments easily

timmarkhuff (approved):
LGTM. I left one comment about how the job is scheduled that I think would be good to consider before merging.

f-wright (approved):
LGTM!

app/status_monitor/status_web.py
CoreyEWood: That seems reasonable, making the switch to the cronjob.

test/escalation_queue/test_queue_retention.py
CoreyEWood: Added!

app/escalation_queue/queue_reader.py
CoreyEWood: Nice, fixed.

app/escalation_queue/manage_reader.py
CoreyEWood: To me it makes sense for this to be lower severity than the one being logged at error because nothing really went wrong here, it's just noting that we're doing a deliberate drop of expired data. So maybe it should actually be info (though it might be worth highlighting it a bit more in case we want to be aware of this occurring). Whereas the error level one is actually something we'd want to look at and take action on.

app/escalation_queue/manage_reader.py
CoreyEWood: I'm not sure if there is, this has been here since Tim added the failed escalation recording. I'm not even sure if we still plan on integrating with Splunk here or not, so might be something to follow up on separately.

PR #451, opened 2026-07-08
feat: add cert-rotation sidecar for in-place TLS cert renewal (GL-1699)

timmarkhuff:
### Summary

Follows up on #448, which pinned the self-signed Lens↔edge TLS cert to device storage so it survives pod restarts. That fixed cert churn on every boot, but the expiry check only ran in the `generate-tls-cert` initContainer at pod startup. A pod running continuously for a year could eventually serve an expired cert.

This PR adds a `cert-rotation` sidecar that checks daily, regenerates the cert if needed, and reloads nginx in place with no pod restart.

### Changes

- **`app/bin/rotate-tls-cert.sh`** (new): daily loop that calls the existing idempotent `generate-tls-cert.sh`, then signals nginx to reload.
- **Helm + legacy k3s manifests**: add `shareProcessNamespace: true` and the `cert-rotation` sidecar.
- **`nginx.conf`**: write `pid /etc/nginx/certs/nginx.pid` so the sidecar can reload the correct nginx in Helm deployments.

### Reload behavior

- **Helm:** real nginx writes its PID to the shared certs volume; sidecar reads it and sends `SIGHUP`.
- **Legacy k3s:** no separate nginx container, so sidecar finds the single nginx master via `/proc` scan (no `procps`/`pgrep` dependency).

The dummy nginx in the Helm `edge-endpoint` container is intentionally not targeted.

### Notes

- At least one device in the field is still on the legacy k3s path; we cannot deprecate it yet.
- Legacy devices won't get cert rotation until they deploy a manifest that includes this sidecar (and #448 cert pinning).
- A daily `SIGHUP` when nothing changed is harmless.

### Jira

https://taserintl.atlassian.net/browse/GL-1699

### Related

- #448 — pinned the cert to device storage (prerequisite)

### DIFF
--- a/app/bin/generate-tls-cert.sh
+++ b/app/bin/generate-tls-cert.sh
@@ -10,7 +10,7 @@
 # every boot. Regeneration is self-healing: a broken pair is replaced rather than
 # left in place for nginx to choke on later.
 
-set -e
+set -ex
 
 # CERT_DIR is overridable so the generator can be exercised in tests; production
 # deployments mount nginx's cert directory here.
--- a/app/bin/generate-tls-cert.sh
+++ b/app/bin/generate-tls-cert.sh
@@ -10,7 +10,7 @@
 # every boot. Regeneration is self-healing: a broken pair is replaced rather than
 # left in place for nginx to choke on later.
 
-set -ex
+set -e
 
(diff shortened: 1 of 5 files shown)


timmarkhuff:
Not needed

PR #453, opened 2026-07-09
fix(install-k3s): retry upstream k3s installer on transient download failures

sunildkumar:
## What & why

`deploy/bin/install-k3s.sh` invokes the upstream `get.k3s.io` installer, which downloads the k3s binary and its checksum from GitHub's release CDN. That download **occasionally fails transiently** — no code is wrong, the CDN just hiccups. Until now a single failure aborted the entire install on the first try.

A recent **zuuul** CI run (`test-k3s-helm`, which checks out `edge-endpoint@main` and runs this exact script) died exactly here:

```
[INFO]  Downloading hash https://github.com/k3s-io/k3s/releases/download/v1.36.2%2Bk3s1/sha256sum-amd64.txt
[ERROR]  Download failed
##[error]Process completed with exit code 1.
```

This script is the **single source of truth** for k3s installs on real edge devices, in edge-endpoint's own CI, and in zuuul's CI. Fixing it here hardens every consumer at once. (Per the task, the zuuul copy of the workflow is a temporary mirror slated for deletion and is intentionally left untouched.)

## Changes

- **New generic `retry <max> <base_sleep> -- cmd…` helper** with linear backoff (10s, then 20s). Factored out so it's unit-testable in isolation.
- **Installer invocation wrapped in `retry 3 …`** (3 attempts). Between attempts a guarded `k3s-uninstall.sh` tears down any partial install so a leftover binary can't make a retry wrongly short-circuit as "already installed".
- **Clear failure path**: if all attempts fail, print an actionable error and exit non-zero.
- **New network-free unit test** `deploy/bin/test/test-install-k3s-retry.sh`. It sources the script in library-only mode (`INSTALL_K3S_LIB_ONLY=1`) and exercises `retry()` against a `PATH`-stubbed fake command:
  1. **Succeeds after transient failures** — fake fails its first N-1 calls, succeeds on the Nth; asserts `retry` returns 0 and called the command exactly N times.
  2. **Gives up after max attempts** — fake always fails; asserts `retry` returns non-zero and stopped after exactly `max` attempts.

## Preserved behavior

All existing behavior is untouched: cgroup checks, cpu/gpu/jetson argument handling, the exact install command and its `K3S_KUBECONFIG_MODE`/`--disable=traefik` args, the k3s version/channel (still `stable`), the "k3s is not installed. Installing…" path, and the remote `curl … | bash -s cpu` install path (the source-only guard is opt-in via an env var that normal execution never sets, so the script stays fully self-contained for `curl | bash`).

## Test output

```
$ ./deploy/bin/test/test-install-k3s-retry.sh
PASS [succeeds-after-transient]: rc=0 after 3 calls
PASS [gives-up]: rc=1 after exactly 3 attempts
All tests passed.
```

### DIFF
--- a/deploy/bin/install-k3s.sh
+++ b/deploy/bin/install-k3s.sh
@@ -5,6 +5,22 @@ set -e
 
 K="k3s kubectl"
 
+# retry MAX BASE -- cmd... : run cmd up to MAX times with linear backoff
+# (BASE, 2*BASE, ... seconds). Returns cmd's exit code from the final attempt.
+retry() {
+    local max=$1 base=$2 attempt=1 rc=0; shift 2; [ "${1:-}" = "--" ] && shift
+    while :; do
+        if "$@"; then return 0; else rc=$?; fi   # capture rc in else, not after the if
+        [ "$attempt" -ge "$max" ] && { echo "retry: '$*' failed after $max attempts (exit $rc)" >&2; return "$rc"; }
+        echo "retry: attempt $attempt/$max failed (exit $rc); retrying in $((base * attempt))s" >&2
+        sleep "$((base * attempt))"
+        attempt=$((attempt + 1))
+    done
+}
+
+# Sourced by the retry unit test (INSTALL_K3S_LIB_ONLY=1) to load helpers only.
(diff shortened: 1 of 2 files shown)


honeytung (approved):
LGTM!

PR #465, opened 2026-07-27
Be specific about the source for `typing-extensions` to keep uv happy

tomfaulhaber:
If I were claude, I would write several paragraphs about this one line change. Suffice to say that `uv run` was breaking and now it's not.


honeytung (approved):
LGTM!

PR #466, opened 2026-07-28
Upgrade pytest to keep dependabot happy

tomfaulhaber:
This required removing a deprecated `pytest.mark` per https://docs.pytest.org/en/stable/deprecations.html#applying-a-mark-to-a-fixture-function


test/api/test_image_queries_live.py
```diff
@@ -24,7 +24,6 @@
 DETECTOR_ID = os.getenv("DETECTOR_ID", "det_2SagpFUrs83cbMZsap5hZzRjZw4")
 
 
-@pytest.mark.live
```
tomfaulhaber, line 27: The pytest documentation claims that `mark` before `fixture` has always been a no-op. It was deprecated for several releases and now causes an error.

honeytung: Yea sounds like this became an error after `9.0`.

honeytung (approved):
LGTM!

